PR #1, opened 2019-08-16
removed all trivial unsafe

djugei:
The reason i would want to use a decoder library in rust is for safety. This library uses a lot of unsafe. I removed some that were trivial to fix (and partially instantly UB).
Others i can not simply remove because its too deeply interwoven with other code, but to me seem to be UB. I added FIXME annotations to those.
Lastly there is some that look like they are fine, but are lacking comments on why they are actually safe. I added TODO annotations to those.

The main use of unsafe was to skip bounds checking on indexing and initializing arrays. Initializing 2 bytes big arrays on the stack costs literally nothing. Indexing can sometimes be avoided by using iterator (added some TODO notes there, too). In other cases please describe why it is actually safe to use unchecked indexing when you have to (after benchmarking ofc). unsafe is not a go faster button and should be handled with a lot of care because it potentially breaks rusts core memory safety promises.

Introduced a dependency on [byteorder](https://docs.rs/byteorder/1.3.2/byteorder/) for endiannes-respecting reading of floats without unsafe casting. you might want to use that in more places, replacing some custom code.

[chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) and [step_by](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.step_by) can be useful for strided iteration.

some of the manual buffer management might possibly be able to utilize [Cursor](https://doc.rust-lang.org/std/io/struct.Cursor.html), not 100% sure about that though.

### DIFF
--- a/sonata-core/src/audio.rs
+++ b/sonata-core/src/audio.rs
@@ -558,6 +558,7 @@ impl<S: Sample> Signal<S> for AudioBuffer<S> {
         assert!(second_idx <self.buf.len());
 
         //FIXME:  this is instant UB, just call chan_pair_mut(0,0) and you get mutable aliasses
+        //maybe try Slice::split_at_mut()
         unsafe {
             let ptr = self.buf.as_mut_ptr();
             (slice::from_raw_parts_mut(ptr.add(first_idx), self.n_frames),
--- a/sonata-core/src/audio.rs
+++ b/sonata-core/src/audio.rs
@@ -308,9 +308,9 @@ impl<S : Sample> AudioBuffer<S> {
         // Practically speaking, it is not possible to allocate more than usize samples.
         debug_assert!(n_sample_capacity <= usize::max_value() as u64);
 
-        // Allocate memory for the sample data, but do not zero the memory.
-        let mut buf = Vec::with_capacity(n_sample_capacity as usize);
-        unsafe { buf.set_len(n_sample_capacity as usize) };
+        // Allocate memory for the sample data, but zero initialize it cause uninitialized memory
+        // is ub pretty much automatically
+        let mut buf = vec![S::default(); n_sample_capacity as usize];
 
         AudioBuffer {
             buf,
@@ -354,6 +354,7 @@ impl<S : Sample> AudioBuffer<S> {
     /// `chan()` to selectively choose the plane to read.
     pub fn planes<'a>(&'a self) -> AudioPlanes<'a, S> {
         let mut planes = AudioPlanes {
+            // FIXME: this is UB
             planes: unsafe { std::mem::uninitialized() },
             n_planes: self.spec.channels.len(),
         };
@@ -374,6 +375,7 @@ impl<S : Sample> AudioBuffer<S> {
     /// `render()`, `fill()`, `chan_mut()`, and `chan_pair_mut()` to mutate the buffer.
     pub fn planes_mut<'a>(&'a mut self) -> AudioPlanesMut<'a, S> {
         let mut planes = AudioPlanesMut {
+            // FIXME: this is UB
             planes: unsafe { std::mem::uninitialized() },
             n_planes: self.spec.channels.len(),
         };
@@ -383,6 +385,7 @@ impl<S : Sample> AudioBuffer<S> {
 
             // Only fill the planes array up to the number of channels.
             for i in 0..planes.n_planes {
+                // FIXME: UB, indexing into uninitialized memory
                 planes.planes[i] = slice::from_raw_parts_mut(ptr as *mut S, self.n_frames);
                 ptr = ptr.add(self.n_capacity);
             }
@@ -554,6 +557,7 @@ impl<S: Sample> Signal<S> for AudioBuffer<S> {
         assert!(first_idx < self.buf.len());
         assert!(second_idx <self.buf.len());
 
+        //FIXME:  this is instant UB, just call chan_pair_mut(0,0) and you get mutable aliasses
         unsafe {
             let ptr = self.buf.as_mut_ptr();
             (slice::from_raw_parts_mut(ptr.add(first_idx), self.n_frames),
@@ -588,6 +592,7 @@ impl<S: Sample> Signal<S> for AudioBuffer<S> {
 
             // Only fill the planes array up to the number of channels.
             for i in 0..planes.n_planes {
+                //FIXME: this is instant UB, you are indexing into uninitialized memory
                 planes.planes[i] = slice::from_raw_parts_mut(ptr as *mut S, n_render_frames);
                 ptr = ptr.add(self.n_capacity);
             }
@@ -607,6 +612,7 @@ impl<S: Sample> Signal<S> for AudioBuffer<S> {
         F: Fn(S) -> S
     {
         debug_assert!(self.n_frames <= self.n_capacity);
+        //TODO: document why this is actually safe
 
         unsafe {
             let mut plane_start = self.buf.as_mut_ptr();
@@ -796,9 +802,10 @@ impl<S: Sample + WriteSample> SampleBuffer<S> {
                 let stride = src.n_capacity;
 
                 for i in 0..n_frames {
+                    //TODO: possibly replace by Slice::chunks() and Iterator::step_by()
                     for ch in 0..n_channels {
-                        let sample = unsafe { src.buf.get_unchecked(ch * stride + i) };
-                        S::write((*sample).into_sample(), &mut writer);
+                        let sample = src.buf[ch * stride + i];
+                        S::write((sample).into_sample(), &mut writer);
                     }
                 }
             },
@@ -843,8 +850,9 @@ impl<S: Sample + WriteSample> SampleBuffer<S> {
                 let stride = src.n_capacity;
 
                 for i in 0..n_frames {
+                    //TODO: possibly replace by Slice::chunks() and Iterator::step_by()
                     for ch in 0..n_channels {
-                        unsafe { S::write(*src.buf.get_unchecked(ch * stride + i), &mut writer); }
+                        S::write(src.buf[ch * stride + i], &mut writer);
                     }
                 }
             },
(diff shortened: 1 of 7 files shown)


terrorfisch:
Seems like AudioPlanesMut and AudioPlanes can be implemented via [`arrayvec`](https://docs.rs/arrayvec/0.4.11/arrayvec/struct.ArrayVec.html) to get rid of the UB.

sonata-core/src/io/media_source_stream.rs
```diff
@@ -281,23 +281,20 @@ impl Bytestream for MediaSourceStream {
...
+        let mut double_byte: [u8; 2] = [0u8; 2];
 
         // If the buffer has two bytes available, copy directly from it and skip any safety or
         // buffering checks.
         if self.pos + 2 < self.end_pos {
-            unsafe { 
-                double_byte[0] = *self.buf.get_unchecked(self.pos + 0);
-                double_byte[1] = *self.buf.get_unchecked(self.pos + 1);
-            }
+            double_byte.clone_from_slice(&self.buf[self.pos..self.pos + 1]);
```
pdeljanov: Should be `&self.buf[self.pos..self.pos + 2]` (+2 rather than +1).

sonata-core/src/io/mod.rs
```diff
@@ -5,6 +5,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use byteorder::{BigEndian, ByteOrder, LittleEndian};
```
pdeljanov: Bringing in byteorder is probably not worthwhile for this one little thing. Please see my general comments about dependencies.

sonata-core/src/audio.rs
```diff
@@ -554,6 +557,7 @@ impl<S: Sample> Signal<S> for AudioBuffer<S> {
         assert!(first_idx < self.buf.len());
         assert!(second_idx <self.buf.len());
 
+        //FIXME:  this is instant UB, just call chan_pair_mut(0,0) and you get mutable aliasses
```
pdeljanov: Amazing catch on this one!

pdeljanov:
Hey djugei, thanks for going over the unsafe code blocks.

I think most of the changes are welcome. I did a quick benchmark on some IO-bound files and there didn't seem to be any performance difference in debug or release mode compared to what's in master right now.

A recurring use-case of unsafe in Sonata is to leave an array uninitialized and then progressively fill it. For example in `read_quad_bytes` or `AudioPlanesMut`. 

In the former example, I'm having a difficult time grasping what makes this undefined behaviour dangerous. Since the array is composed of `u8`s, I'm not sure why initializing each value one-by-one is bad. Am I missing something here?

Regarding the second example with `AudioPlanes` and `AudioPlanesMut`, this is actually a part of the API that I find pretty ugly. The intention was to allow access to an entire frame of audio at a time. Since there are up to 32 audio planes (channels), I fill an array with references to each available plane, and then only return a slice of the buffer containing valid references. As terrorfisch mentioned, `arrayvec` is a good alternative, but overall the API is still inelegant. This is only used in the FLAC decoder's output audio validation routine. If a different method is used there, perhaps by using a `SampleBuffer`, then this whole part of the API can be scraped.

> Introduced a dependency on byteorder for endiannes-respecting reading of floats without unsafe casting. you might want to use that in more places, replacing some custom code.

I don't think it's really worth it to bring in byteorder simply for doing the byte swaps required for f32 and f64. The original, albeit, unsafe code is actually pretty much exactly what byteorder itself does. Since all the other integer types are handled by the standard library, I think it's acceptable to just do it manually for now. I don't expect to encounter many more endian issues going forward.

I'd like Sonata to use as few dependencies as possible. This isn't a hard rule, there's flexibility on this of course, but for small things or things that would do better with a more specialized approach, I don't think it's worth it. 

I did leave some comments on the PR though because there were a few minor errors that broke a few things and would have to be fixed before merging the request.

Thanks!

sonata-core/src/sample.rs
```diff
@@ -34,6 +34,7 @@ pub enum SampleFormat {
 /// underlying data type. Additionally, `Sample` provides information regarding the
 /// format of underlying data types representing the sample when in memory, but also
 /// when exported.
+pub trait Sample: Copy + Clone + Sized + Default {
 pub trait Sample : Copy + Clone + Sized {
```
pdeljanov: This line should be deleted if replaced by the previous line, it breaks the build.

djugei: right, sorry i did a line-by-line commit because rustfmt changed a lot of unrelated code that i did not want to muddle the pr with. gonna fix that

sonata-core/src/io/mod.rs
djugei: byteorder is probably able to replace most code in that file, basically all the integer-parsing would just shell out to byteorder

djugei:
> A recurring use-case of unsafe in Sonata is to leave an array uninitialized and then progressively fill it. For example in `read_quad_bytes` or `AudioPlanesMut`.
> In the former example, I'm having a difficult time grasping what makes this undefined behaviour dangerous. Since the array is composed of `u8`s, I'm not sure why initializing each value one-by-one is bad. Am I missing something here?

yes, as it states on the (documentation for uninitalized)[https://doc.rust-lang.org/std/mem/fn.uninitialized.html] basically just using this at all is a problem, even on integer types.

for the progressively filling array on the stack you can use arrayvec, as mentioned by @terrorfisch .
that would also save a lot of error prone custom code with offsets'n stuff.

for performance it is just not worth it.

> > Introduced a dependency on byteorder for endiannes-respecting reading of floats without unsafe casting. you might want to use that in more places, replacing some custom code.
> 
> I don't think it's really worth it to bring in byteorder simply for doing the byte swaps required for f32 and f64. The original, albeit, unsafe code is actually pretty much exactly what byteorder itself does. Since all the other integer types are handled by the standard library, I think it's acceptable to just do it manually for now. I don't expect to encounter many more endian issues going forward.

byteorder also has the advantage of being able to read directly from anything implementing Read, letting you get rid of quite some manual buffer management.

moving unsafetly out of your own crate allows you to declare #![deny(unsafe)] at the top of your crate, which makes everybody more happy and introduces another (mental) barrier against introducing unsafe in other places again.

> 
> I'd like Sonata to use as few dependencies as possible. This isn't a hard rule, there's flexibility on this of course, but for small things or things that would do better with a more specialized approach, I don't think it's worth it.

I think this is a bit of a C++ mindset, cause i know its a huge fucking pain to add dependencies there. adding a dependency carries almost no cost thanks to cargo. ofc it is important to check what you depend on for trustworthyness, but byteorder and arrayvec are kinda well-known much relied on crates, so there is no reason to be too worried. try re-evaluating what reasons you have for trying to have minimal dependencies and see if they still apply.

specifically for byteorder, you can basically replace the whole Bytestream Trait by byteorders (ByteOrder)[https://docs.rs/byteorder/1.3.2/byteorder/trait.ByteOrder.html] if im not mistaken.
This is also a bit of an interoperability thing, as ppl are probably more used to consuming ByteOrder and Read as opposed to an almost identical custom trait.

> I did leave some comments on the PR though because there were a few minor errors that broke a few things and would have to be fixed before merging the request.
yup gonna fix those in 10 minutes.

pdeljanov:
> > > Introduced a dependency on byteorder for endiannes-respecting reading of floats without unsafe casting. you might want to use that in more places, replacing some custom code.
> > 
> > 
> > I don't think it's really worth it to bring in byteorder simply for doing the byte swaps required for f32 and f64. The original, albeit, unsafe code is actually pretty much exactly what byteorder itself does. Since all the other integer types are handled by the standard library, I think it's acceptable to just do it manually for now. I don't expect to encounter many more endian issues going forward.
> 
> byteorder also has the advantage of being able to read directly from anything implementing Read, letting you get rid of quite some manual buffer management.
> 
> moving unsafetly out of your own crate allows you to declare #![deny(unsafe)] at the top of your crate, which makes everybody more happy and introduces another (mental) barrier against introducing unsafe in other places again.
> 
> > I'd like Sonata to use as few dependencies as possible. This isn't a hard rule, there's flexibility on this of course, but for small things or things that would do better with a more specialized approach, I don't think it's worth it.
> 
> I think this is a bit of a C++ mindset, cause i know its a huge fucking pain to add dependencies there. adding a dependency carries almost no cost thanks to cargo. ofc it is important to check what you depend on for trustworthyness, but byteorder and arrayvec are kinda well-known much relied on crates, so there is no reason to be too worried. try re-evaluating what reasons you have for trying to have minimal dependencies and see if they still apply.
> 
> specifically for byteorder, you can basically replace the whole Bytestream Trait by byteorders [ByteOrder](https://docs.rs/byteorder/1.3.2/byteorder/trait.ByteOrder.html) if im not mistaken.
> This is also a bit of an interoperability thing, as ppl are probably more used to consuming ByteOrder and Read as opposed to an almost identical custom trait.
> 
> > I did leave some comments on the PR though because there were a few minor errors that broke a few things and would have to be fixed before merging the request.
> 
> yup gonna fix those in 10 minutes.

Pretty solid arguments. It looks like byteorder's `ReadBytesExt` is close to Sonata's `Bytestream`. Since `MediaSourceStream` implemented `Read` then it should be possible to use `ReadBytesExt` pretty easily. I'm not certain at this point that Sonata's `Bytestream` trait yet, but it warrants investigation, so I'll make an Issue. We do have a few extra functions such as `scan_bytes`, `ignore_bytes`, and `read_boxed_slice_bytes` which are very useful. So maybe extending `ReadBytesExt` is the best way to go.

Also, the relative performance of `read_u8` will be critical. Sonata's bit reader consumes a byte at a time, so `read_u8` is in the hot path. I see that byteorder depends on `read_exact` to get data, which is a more complex function than the bounds check and copy the current `Bytestream` performs. Since bit stream performance is critical, this may be a major problem in adopting byteorder.

Aside from that, thank you for making the changes, I'll be merging your PR. Cheers!

djugei:
sweet

PR #2, opened 2019-08-18
Fix build on non-linux

terrorfisch:
```rust
#[cfg(not(target_os = "linux"))]
fn play(_: Box<dyn FormatReader>, _: &DecoderOptions) -> Result<()> {
    // TODO: Support the platform.
    unsupported_error("Playback is not supported on your platform.")
}
```
Did not compile.


pdeljanov:
Thanks for the fix!

Issue #6, opened 2019-09-17
Files used for benchmarks are not published


Shnatsel:
In absence of files used for the benchmarks listed in README.md it's impossible to reproduce the performance measurement results. This hinders attempts to improve performance further.

Please share the files used for the measurements.

Shnatsel:
The reason I'm interested is that I recall that `codegen-units = 1` annotation in Cargo.toml sped up Claxon a great deal. I was wondering if this would be the case for Sonata as well, and also to test it against my local ffmpeg. Sadly I am unable to build it due to a bug in Pulseaudio (http://pad.lv/1796324) so I guess the point is moot.

pdeljanov:
> In absence of files used for the benchmarks listed in README.md it's impossible to reproduce the performance measurement results. This hinders attempts to improve performance further.
> 
> Please share the files used for the measurements.

Unfortunately I just used random files from my music library. Since they are copyrighted I wouldn't be able to share them. Are there any Creative Commons licensed sets available?

I found that for FLAC specifically the bit-depth and sampling rate (thus bit-rate) affected the performance more so than the difference between similarly encoded files. This makes sense, for Sonata, about 75% of the total CPU time for FLAC is spent in the bitstream reader.

> The reason I'm interested is that I recall that `codegen-units = 1` annotation in Cargo.toml sped up Claxon a great deal. I was wondering if this would be the case for Sonata as well, and also to test it against my local ffmpeg. Sadly I am unable to build it due to a bug in Pulseaudio (http://pad.lv/1796324) so I guess the point is moot.

I had that bug too when compiling Sonata on my BananaPi. I solved it by upgrading to the latest PulseAudio. If you don't care about playback, in [sonata-play/src/main.rs](https://github.com/pdeljanov/Sonata/blob/master/sonata-play/src/main.rs) you can simply delete the Linux-specific playback code. The changes are minimal, simply remove any functions or statements with the `#[cfg(target_os = "linux")]` attribute. And remove any `#[cfg(not(target_os = "linux"))]` attributes elsewhere.

I've tried `codegen-units = 1` before and it never made a difference. Maybe I did it wrong. My impression is that Sonata's main bottleneck is the bitreader. It's fast, but since it pulls 1 byte from the stream at a time it can become a problem on certain files. I have a private branch that should make it many times faster, but it's a tricky thing to get right and not a big priority at the moment.

Shnatsel:
[...]

PR #7, opened 2020-04-07
Script to run benchmark tests.

richardmitic:
A selection of files in mp3, flac, and wav format are downloaded
from archive.org and benchmarking is perfomed by hyperfine.

Fixes issue #6


pdeljanov:
Thanks for this!

PR #12, opened 2020-08-01
ReadSeekSource as MediaSource

AmionSky:
Like ReadOnlySource but allows Seeking.
It allows creating a MediaSource from a generic type with support for most of the features.

`MediaSource::len()` could be updated with `std::io::Seek::stream_len()` when it gets stabilized if it's desired.

### DIFF
--- a/sonata-core/src/io/mod.rs
+++ b/sonata-core/src/io/mod.rs
@@ -75,6 +75,57 @@ impl<T: std::convert::AsRef<[u8]>> MediaSource for io::Cursor<T> {
     }
 }
 
+/// `ReadSeekSource` implements a seekable `MediaSource` for any reader that implements the
+/// `std::io::Read` and `std::io::Seek` traits.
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hi @AmionSky,

Unfortunately I don't think this is pull request is necessary. The motivation for `ReadOnlySource` is to wrap readers that do not implement `std::io::Seek` (for example, `std::io::Stdin`). For readers that implement both `std::io::Read` and `std::io::Seek` a user of the library simply has to `impl MediaSource` on the reader. Sonata already does for `std::fs::File` and `std::io::Cursor` for the file and in-memory buffer cases.

Is there a use-case I'm missing?

AmionSky:
It's true that we can just implement MediaSource on something for our use case (I could just move the entire ReadSeekSource into my project) but the existence of ReadOnlySource suggested that it was desired to have generic approaches in the core lib.

PR #21, opened 2021-03-17
Change AudioPlanes::planes() to &self instead of &mut self

Timmmm:
Pretty sure it doesn't need &mut.


pdeljanov:
Good catch, looks like a copy and paste error. Thanks for the fix!

PR #22, opened 2021-03-22
Add GitHub actions workflow for testing builds

### LINKED ISSUE #11: Configure CI
Configuring CI would have caught the compilation issue I resolved in #10. I don't care whether the project uses GitHub Actions or another system, but I would strongly recommend setting up a minimal pipeline to at least run tests on every commit.

I can easily configure GitHub Actions if this is something the project would like to see done.

udoprog:
This will build and test pull requests and the master branch when pushed to.

Note that `rustfmt` check is included, but since codebase is not formatted it is currently commented out. It would however be nice to format the code and include it, since manual formatting is a bit tedious and arbitrary.

You can see an example build of this here: https://github.com/udoprog/Symphonia/actions/runs/674345665

Fixes #11

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,39 @@
+name: Build and Test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request: {}
+
+jobs:
+  build:
+    name: Build
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Setup Toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+# Uncomment to check formatting:
+#    - name: Check Formatting
+#      uses: actions-rs/cargo@v1
+#      with:
+#        command: fmt
+#        args: -- --check
+    - name: Build
+      uses: actions-rs/cargo@v1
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks for this, it's great. I never explored GitHub's CI features before.

I'll revisit rustfmt and see what can be done about the formatting. The reason why there isn't a rustfmt configuration for this project is because the last time I tried I found that it mangled things a bit too much, such as alignment of the elements in a large lookup table.

udoprog:
> I'll revisit rustfmt and see what can be done about the formatting. The reason why there isn't a rustfmt configuration for this project is because the last time I tried I found that it mangled things a bit too much, such as alignment of the elements in a large lookup table.

You can add `#[rustfmt::skip]` when referencing any generated modules, like:

```rust
#[rustfmt::skip]
mod huffman_tables;
```

And for mixed files, the pattern I'd recommend for including large arrays of raw data in projects would be to put it in a dedicated file and `include!` it. It helps to keep data and code separate. E.g. `crc32.rs`:

```rust
include!("crc32_table.rs");
```

The rest I'd just ignore or get used to, Formatting code selectively or by hand is such a bigger hassle compared to a few things being offensively formatted.

pdeljanov:
Thanks, those are great suggestions. I didn't know about the `include!` macro, that'll actually tidy-up things like the checksum module a lot.

I've merged your workflow as-is into the repository. When I have more time to work on Symphonia I will be sure to try your suggestions. Cheers!

PR #23, opened 2021-03-22
Add support for a few more atoms and components required for m4a files

udoprog:
This removes the requirements that FourCC codes have to be readable ASCII and adds another track type `b"text"` to hdlr.

I encountered an `.m4a` file that reported its major brand as b"m4a\0" which also included this track type. I think It's legal according to the spec, but I'd have to double check this.


pdeljanov:
Thanks for your fixes! Indeed, I checked the standard, and it seems to say that while FourCCs _should_ be printable characters, the four characters aren't actually restricted to printable ASCII characters.

PR #25, opened 2021-04-10
Fix #24: Require MediaSource to be Send

### LINKED ISSUE #24: MediaSourceStream is not Send
Due to wrapping a `Box<dyn MediaSource>` instead of `Box<dyn MediaSource + Send>`, MediaSourceStream is not safe to send across thread boundaries. Is this intentional?

james7132:
This should allow `MediaSource` based types like `MediaSourceStream` to be Send. Otherwise, as it stands, these streams cannot be sent from one thread to another. With most audio streams, it should be a safe assumption that they can be moved from thread to thread, but likely not `Sync`.


pdeljanov:
This one ended up getting addressed by #31. Thank you for the PR though!

PR #26, opened 2021-04-10
Fix UnexpectedEof when using Read::read_to_end on MediaSourceStream

james7132:
Calling `Read::read_to_end` on a `MediaSourceStream` returns an UnexpectedEof instead of properly returning the read values. This is  particularly useful in unit tests. This PR fixes this issue.

Added unit tests to ensure this remains fixed.

### DIFF
--- a/symphonia-core/src/io/media_source_stream.rs
+++ b/symphonia-core/src/io/media_source_stream.rs
@@ -255,8 +255,6 @@ impl MediaSource for MediaSourceStream {
 impl io::Read for MediaSourceStream {
 
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let buf_len = buf.len();
-
         let mut rem = buf;
         let mut read = 0;
 
@@ -278,13 +276,7 @@ impl io::Read for MediaSourceStream {
             }
         }
 
-        // Return an end-of-stream error if 0 bytes were read when atleast 1 byte was requested.
-        if read == 0 && buf_len > 0 {
-            Err(io::Error::new(io::ErrorKind::UnexpectedEof, END_OF_STREAM_ERROR_STR))
-        }
-        else {
-            Ok(read)
-        }
+        Ok(read)
     }
 
 }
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hi @james7132, thanks for this.

It's a good catch, `Read` trait implementations should return 0 on EOF conditions. However, with the current PR there will be many regressions. This is because the format readers all use the `ByteStream` trait of `MediaSourceStream`, and the semantics of this trait is such that on an EOF condition an error is returned, not 0. Since the implementation of `ByteStream::read_buf` for `MediaSourceStream` simply calls `Read::read`, many format readers will just spin in a loop at the end of a stream trying to read more.

Luckily, this is easy to fix. Just move the error checking logic from the end of the `Read::read` implementation to the `ByteStream::read_buf` implementation for `MediaSourceStream`.

I believe this should fix the regressions since the `Read` trait is not typically used in Symphonia.

symphonia-core/src/io/media_source_stream.rs
```diff
@@ -278,13 +276,7 @@ impl io::Read for MediaSourceStream {
             }
         }
 
-        // Return an end-of-stream error if 0 bytes were read when atleast 1 byte was requested.
```
pdeljanov, line 281: This logic should be moved to the `ByteStream::read_buf` implementation of `MediaSourceStream` to keep the semantics of the `ByteStream` trait the same.

symphonia-core/src/io/media_source_stream.rs
```diff
@@ -278,13 +276,7 @@ impl io::Read for MediaSourceStream {
...
-        if read == 0 && buf_len > 0 {
-            Err(io::Error::new(io::ErrorKind::UnexpectedEof, END_OF_STREAM_ERROR_STR))
-        }
-        else {
-            Ok(read)
-        }
```
pdeljanov, line 287: This logic should be moved to the `ByteStream::read_buf` implementation of `MediaSourceStream` to maintain the existing semantics of `ByteStream`.

james7132: Done.

pdeljanov (approved):

pdeljanov:
Thanks, merged!

PR #29, opened 2021-05-30
Fix symphonia-play compilation when using cpal

Herschel:
Add `buffer_size` parameter to `cpal::StreamConfig`.


pdeljanov:
Thanks for catching this!

PR #30, opened 2021-06-09
Add into_inner function to allow access to the stream

aschey:
This adds a way to destroy the `FormatReader` and get access to the underlying stream. This is useful if you need to directly manipulate the stream and then restart playback. I'm working on integrating Symphonia with Rodio and this technique is used [here](https://github.com/RustAudio/rodio/blob/master/src/decoder/mod.rs#L262) to loop an audio source.

### DIFF
--- a/symphonia-bundle-flac/src/demuxer.rs
+++ b/symphonia-bundle-flac/src/demuxer.rs
@@ -251,6 +251,10 @@ impl FormatReader for FlacReader {
         }
     }
 
+    fn into_inner(self: Box<Self>) -> MediaSourceStream {
+        self.reader
+    }
+
 }
 
 /// Reads a StreamInfo block and populates the reader with stream information.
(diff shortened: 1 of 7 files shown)


pdeljanov (changes requested):
This is a good idea, but I think `MetadataReader` should also provide `into_inner` to serve a similar purpose.

aschey:
> This is a good idea, but I think `MetadataReader` should also provide `into_inner` to serve a similar purpose.

How should that work exactly? Maybe I'm missing something, but I only see `Id3v2Reader` that implements `MetadataReader` and it doesn't look like that struct holds a reference to anything.

pdeljanov:
> > This is a good idea, but I think `MetadataReader` should also provide `into_inner` to serve a similar purpose.
> 
> How should that work exactly? Maybe I'm missing something, but I only see `Id3v2Reader` that implements `MetadataReader` and it doesn't look like that struct holds a reference to anything.

Sorry, you're right, `MetadataReader` does not take ownership of the stream. I should've checked instead of relying on my memory. :laughing:

aschey:
No worries, thanks!

PR #31, opened 2021-06-09
Require FormatReader and Decoder to be Send

### LINKED ISSUE #24: MediaSourceStream is not Send

aschey:
Resolves #24 

Similar to https://github.com/pdeljanov/Symphonia/pull/25, but this also requires `FormatReader` and `Decoder` to be `Send`.  The requirements for the other traits to be `Send` are all downstream effects of this. Per https://github.com/pdeljanov/Symphonia/issues/24#issuecomment-817185812, I need to keep a reference to the format reader and the decoder in a struct in order to keep playing the stream. The only other change here is switching an `Rc` to an `Arc`.

### DIFF
--- a/symphonia-format-isomp4/src/demuxer.rs
+++ b/symphonia-format-isomp4/src/demuxer.rs
@@ -16,8 +16,8 @@ use symphonia_core::meta::MetadataQueue;
 use symphonia_core::probe::{Descriptor, Instantiate, QueryDescriptor};
 use symphonia_core::units::Time;
 
-use std::rc::Rc;
 use std::io::{Seek, SeekFrom};
+use std::sync::Arc;
 
 use crate::atoms::{AtomIterator, AtomType};
 use crate::atoms::{FtypAtom, MoovAtom, MoofAtom, SidxAtom, TrakAtom, MetaAtom, MvexAtom};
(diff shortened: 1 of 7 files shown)


symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -405,7 +405,7 @@ impl FormatReader for IsoMp4Reader {
 
         // A Movie Extends (mvex) atom is required to support segmented streams. If the mvex atom is
         // present, wrap it in an Rc so it can be shared amongst all segments.
```
pdeljanov, line 407: Please update the comment to note it's an Arc instead of Rc.

pdeljanov (changes requested):
I think `MetadataReader` should be `Send` as well even though the use-case is more limited.

symphonia-format-isomp4/src/demuxer.rs
aschey: Good catch, thanks. Fixed.

aschey:
> I think `MetadataReader` should be `Send` as well even though the use-case is more limited.

Makes sense, just pushed that change as well.

PR #33, opened 2021-06-30
Make MetadataQueue's modification methods consistent

BlackHoleFox:
Hiya. When trying to bootstrap up a project with Symphonia I noticed an oddity/two with `MetadataQueue`. Its `pop()` method took an immutable reference, relying on the `RefCell` to get a mutable borrow, but the `push()` method took a mutable reference anyway. It seemed like an odd API disjoint, so I thought I'd try cleaning it up.

I chose to make `pop()` take a `&mut self` as well, to be consistent with most crates and to allow for the internals to be cleaned up. After that bit, there wasn't any need for the `RefCell` at all, so I removed it entirely. That also let me tweak `MetadataRef` to remove a panic, by just taking a regular reference.

In the end, `MetadataRef` seemed like it wasn't worth having anymore so I removed it in a separate commit. Happy to drop that if you think its too much. Unless users are holding these themselves though, there shouldn't be much of a migration path at all due to the previous `Deref` impl.

### DIFF
--- a/symphonia-core/src/meta.rs
+++ b/symphonia-core/src/meta.rs
@@ -8,12 +8,10 @@
 //! The `meta` module defines basic metadata elements, and management structures.
 
 use std::borrow::Cow;
-use std::cell::{Ref, RefCell};
 use std::collections::VecDeque;
 use std::convert::From;
 use std::fmt;
 use std::num::NonZeroU32;
-use std::ops::Deref;
 
 use crate::errors::Result;
 use crate::io::MediaSourceStream;
@@ -402,13 +400,13 @@ pub struct VendorData {
 
 /// `Metadata` is a container for a single discrete revision of metadata information.
 #[derive(Default)]
-pub struct Metadata {
+pub struct MetadataRevision {
     tags: Vec<Tag>,
     visuals: Vec<Visual>,
     vendor_data: Vec<VendorData>,
 }
 
-impl Metadata {
+impl MetadataRevision {
     /// Gets an immutable slice to the `Tag`s in this revision.
     pub fn tags(&self) -> &[Tag] {
         &self.tags
@@ -427,7 +425,7 @@ impl Metadata {
 
 /// `MetadataBuilder` is the builder for `Metadata` revisions.
 pub struct MetadataBuilder {
-    metadata: Metadata,
+    metadata: MetadataRevision,
 }
 
 impl MetadataBuilder {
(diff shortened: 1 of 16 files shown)


pdeljanov:
Hey, thanks for this.

I can see how this looks a little weird, but a little background may help.

As a `FormatReader` reads through a container during playback it may encounter one or more new pieces of metadata. I wanted to delegate the responsibility of how that new metadata is handled to the application. So,  `FormatReader`s will just queue up a  new "revision" onto the `MetadataQueue` and then the application can interrogate the `is_latest()` flag to determine if it should pop the old revision(s) off. Then, the application can decide what to do with this new piece of metadata (merge it into the old, replace the old, etc).

So the reason why the signatures were done this way was because `FormatReader::metadata()` returns an immutable reference to the `MetadataQueue`, and the reason for that was so that an application can't `push()` on metadata onto the queue. The only thing the application should be able to do is pop off one or more revisions.

Do you think this could be improved?

BlackHoleFox:
Thanks again for clarifying here. With that, I can see now why you went for this design. Off the top of my head, putting `#[doc(hidden)]` on `push()` seems like a start. Checking around where its referenced, it only seems to be meant for in-workspace use of symphonia.

However, I still think theres a bit of room for improvement around a more misuse-resistant API. I say that because I could easily imagine someone doing this while handling new metadata that is given to them by `FormatReader`, determining something.

For example, this would panic for the user if they tried to scan the latest metadata in order to determine if they needed other metadata chunks in the queue. I don't think this would be super uncommon either (I was planning on having something similar).
```rust
let metadata_queue = probed.format.metadata();
// This was determined in some way 
let artist_name: &str = { ... };

if let Some(metadata) = metadata_queue.current() {
    // Check if file has duplicate metadata pieces for cleanup.
     if metadata.tags().iter().any(|t| matches!(t.std_key, Some(StandardTagKey::AlbumArtist)) && matches!(t.value, Value::String(ref s) if s == artist_name)) {
          metadata_queue.pop();
     }
}
```

I tried thinking about a way to cleanly hide the `RefCell` from the end user to avoid this, but haven't come up with anything yet. I'll think on it more though and see if something floats around.

pdeljanov:
> Thanks again for clarifying here. With that, I can see now why you went for this design. Off the top of my head, putting `#[doc(hidden)]` on `push()` seems like a start. Checking around where its referenced, it only seems to be meant for in-workspace use of symphonia.

I don't think we should do that because even though all the codecs and formats are in the same workspace they ultimately don't have to be. A design goal is to allow anyone to implement the `Decoder` or `FormatReader` trait and register their own codecs and containers. In that case, we probably shouldn't hide the `push()` documentation since they'd need it.

> However, I still think theres a bit of room for improvement around a more misuse-resistant API. I say that because I could easily imagine someone doing this while handling new metadata that is given to them by `FormatReader`, determining something.
> 
> For example, this would panic for the user if they tried to scan the latest metadata in order to determine if they needed other metadata chunks in the queue. I don't think this would be super uncommon either (I was planning on having something similar).
> 
> ```rust
> let metadata_queue = probed.format.metadata();
> // This was determined in some way 
> let artist_name: &str = { ... };
> 
> if let Some(metadata) = metadata_queue.current() {
>     // Check if file has duplicate metadata pieces for cleanup.
>      if metadata.tags().iter().any(|t| matches!(t.std_key, Some(StandardTagKey::AlbumArtist)) && matches!(t.value, Value::String(ref s) if s == artist_name)) {
>           metadata_queue.pop();
>      }
> }
> ```
> 
> I tried thinking about a way to cleanly hide the `RefCell` from the end user to avoid this, but haven't come up with anything yet. I'll think on it more though and see if something floats around. Please feel free to correct me if anything here seems obviously wrong and something that wouldn't really make sense anyway.

Oh yikes... That's a good example and a big hidden hazard. You're completely correct.

In my mind I had only considered the scenarios below which wouldn't encounter the issue.

```rust
// Case 0: Just querying for metadata.
if let Some(metadata) = reader.metadata().current() {
    // Do something with metadata...
}

// Case 1: Consume all metadata revisions all at once. Probably would do this after
// instantiating a FormatReader.
loop {
    if let Some(metadata) = reader.metadata().pop() {
        // Do something with metadata...
    }
   else {
      break;
   }
}

// Case 2: Check if new metadata is present after reading a packet, and consuming it.
if !reader.metadata().is_latest() {
    if let Some(metadata) = reader.metadata().pop() {
        // Do something with metadata...
    }
}
```

I'll also need to think more deeply about this. Off the top of my head I'm drawing blanks.

BlackHoleFox:
I thought about this for a bit and the only option I could come up with is making `push` take itself by value in a builder-esque way. This is a little bit ugly internally, but it didn't lead to anything unreasonable imo, with the added gain of not needing a lot of deep clones.

symphonia-core/src/meta.rs
```diff
@@ -462,63 +461,44 @@ impl MetadataBuilder {
...
-    pub fn pop(&self) -> Option<Metadata> {
-        let mut queue = self.queue.borrow_mut();
-
-        if queue.len() > 1 {
-            queue.pop_front()
+    pub fn pop(&mut self) -> Option<Metadata> {
+        if self.queue.len() > 1 {
+            self.queue.pop_front().map(|m| match Arc::try_unwrap(m) {
+                Ok(m) => m,
+                Err(_) => unreachable!("no references should exist when queue is borrowed mutably")
```
BlackHoleFox, line 490: i need to look at this again since im not confident in this

BlackHoleFox: e50da23f8c1cefd0fb48e7e28b6b27b67cc1e96b expands on this idea some

pdeljanov:
This is very clever, but it's also a bit unintuitive.

After further thinking, I believe the following is a simpler solution with (hopefully) no footguns that I can see.

Godbolt link with a working example: https://godbolt.org/z/b1Kao6M3j.

```rust
// NOTE: Illustrative only. MetadataRevision is currently called Metadata upstream.
// The original struct should only be renamed.
#[derive(Default)]
struct MetadataRevision {
    tags: Vec<Tag>,
}

struct Metadata<'a> {
    revs: &'a mut VecDeque<MetadataRevision>,
}

impl<'a> Metadata<'a> {
    pub fn is_latest(&self) -> bool {
        self.revs.len() == 1
    }

    pub fn current(&self) -> Option<&MetadataRevision> {
        self.revs.front()
    }

    pub fn pop(&mut self) -> Option<MetadataRevision> {
        if !self.is_latest() {
            self.revs.pop_front()
        }
        else {
            None
        }
    }
}

#[derive(Default)]
struct MetadataLog {
    revs: VecDeque<MetadataRevision>,
}

impl MetadataLog {
    pub fn metadata(&mut self) -> Metadata<'_> {
        Metadata { revs: &mut self.revs }
    }

    pub fn push(&mut self, rev: MetadataRevision) {
        self.revs.push_back(rev);
    }
}

// Updated FormatReader trait.
trait FormatReader : Send {
    // .. snip ..
    
    fn metadata(&mut self) -> Metadata<'_>;
    
    // .. snip ..
}
```

A `FormatReader` will contain a `MetadataLog`. The `MetadataLog` contains a queue of `MetadataRevision`s. The `MetadataLog` provides a `push()` function to add new revisions which the `FormatReader` can freely use. When the `FormatReader::metadata()` function is called, we will return the result of  `MetadataLog::metadata()`. This function returns a `Metadata` which is a wrapper around a mutable reference to the underlying queue. `Metadata` only implements the allowed public functions, and in this way a user can't modify the queue in other ways.

I renamed what's currently called `Metadata` upstream to `MetadataRevision` since that better describes what it is (a collection of tags, visuals, etc. at a certain point in time).

I think this'll resolve all the issues in a tidy way. What do you think?

BlackHoleFox:
> I think this'll resolve all the issues in a tidy way. What do you think?

I agree and I'm sort of embarrassed I came up with my idea instead of just splitting things into multiple types :D 

The only thing I found to be sorta lackluster with this approach is that when combined with the changes #34 made, accessing the "current" metadata with combinators becomes a bit ugly.

pdeljanov:
Hmmm, you're right, the combinators are quite ugly.

Do you think `ProbeResult` should return `Option<Vec<MetadataRevision>>` instead of `Option<MetadataLog>`?

My only concern with that approach is that users could've had a single function for iterating over `Metadata` whereas in this case they'll need two different functions. On the other hand, they can still have a common function to handle each `MetadataRevision` so maybe it's not a big deal. Another thought: maybe `Metadata` could provide a function to get an immutable slice over the `MetadataRevision`s and then in both cases they can iterate over the slice. `Metadata` would then need a way to pop all of the latest revisions off after they've completed the iteration. :thinking:

BlackHoleFox:
It feels like removing the strongly typed structures wouldn't be worth it just to have some easier accesses. I tried out an idea of a specialized helper structure to make combinators. In `symphonia-play`, I was able to get rid of one layer of chaining since the guarantee that if the `ProbeResult->metdata` is `Some(...)`, then the log will have a value in it, so you don't also need to call `.current()` to check in that specific case.

pdeljanov (approved):
Looks good to me!

I like this solution, it really ties everything nicely together. Thanks for sticking with this!

PR #34, opened 2021-06-30
Document when ProbeResult's metadata queue is empty

BlackHoleFox:
When trying to use the library initially, it didn't seem clear that this would be empty all the time. It does make more sense if you read the `symphonia-player` example or look at the source, but it could probably gain from being more forward in the docs too.

Another idea: `ProbeResult` could be made into an enum instead, though this doesn't seem great and the docs are a simpler change since I don't think theres a real difference between the container being at the front of the stream and somewhere later after metadata, but I could be wrong there.
```rust
pub enum ProbeResult {
    InitalContainer {
        /// An instance of a `FormatReader` for the probed format.
        format: Box<dyn FormatReader>,
    },
    Container {
        /// An instance of a `FormatReader` for the probed format.
        format: Box<dyn FormatReader>,
        /// A queue of `Metadata` revisions read during the probe operation before the instantiation of
        /// the `FormatReader`. This queue will contain at least one item.
        metadata: MetadataQueue,
    }
}
```

### DIFF
--- a/symphonia-core/src/probe.rs
+++ b/symphonia-core/src/probe.rs
@@ -161,9 +161,11 @@ pub struct ProbeResult {
     /// An instance of a `FormatReader` for the probed format
     pub format: Box<dyn FormatReader>,
     /// A queue of `Metadata` revisions read during the probe operation before the instantiation of
-    /// the `FormatReader`. If the container format was immediately determined, this will be empty and metadata
-    /// should be obtained by calling `.metadata()` on `format`.
-    pub metadata: MetadataQueue
+    /// the `FormatReader`. If any additional metadata was present outside of the container, this is `Some` and the queue will have
+    /// at least one item in it.
+    ///
+    /// Metadata that was part of the container format itself can be read by calling `.metadata()` on `format`.
+    pub metadata: Option<MetadataQueue>,
 }
 
 /// `Probe` scans a `MediaSourceStream` for metadata and container formats, and provides an
@@ -303,6 +305,12 @@ impl Probe {
                 // If a container format is found, return an instance to it's reader.
                 Instantiate::Format(fmt) => {
                     let format = fmt(mss, format_opts)?;
+
+                    let metadata = if metadata.current().is_some() {
+                        Some(metadata)
+                    } else {
+                        None
+                    };
 
                     return Ok(ProbeResult { format, metadata });
                 }
--- a/symphonia-core/src/probe.rs
+++ b/symphonia-core/src/probe.rs
@@ -161,7 +161,8 @@ pub struct ProbeResult {
     /// An instance of a `FormatReader` for the probed format
     pub format: Box<dyn FormatReader>,
     /// A queue of `Metadata` revisions read during the probe operation before the instantiation of
-    /// the `FormatReader`.
+    /// the `FormatReader`. If the container format was immediately determined, this will be empty and metadata
+    /// should be obtained by calling `.metadata()` on `format`.
     pub metadata: MetadataQueue
 }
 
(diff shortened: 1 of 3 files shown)


pdeljanov:
So, some metadata formats aren't actually embedded into the container itself but are rather just blobs of data tacked onto the beginning or end of a stream. For example, the ID3v2 tags on MP3 files are actually encoded in such a way that a MP3 decoder would ignore it as invalid data. A compatible decoder however will notice the tags and read them.

If Symphonia followed that model, then specific `FormatReader`s would need to implement support for specific metadata formats and this seemed like a bad idea. For example, I have FLAC files with ID3v2 tags on them for some reason.

Instead, Symphonia has `Probe` which scans through a media stream until it encounters a compatible media format and instantiates the proper `FormatReader`. As it scans the stream though, if it encounters any metadata, then it'll instantiate a compatible `MetadataReader` to read the metadata and queue it onto a `MetadataQueue`. Then it'll continue scanning. That `MetadataQueue` is ultimately returned in the `ProbeResult`.

Of course, some streams don't have this tacked on metadata (though apparently some tools add it anyways). Formats like FLAC and M4A embed metadata within the container itself and that's what you get when you look at `FormatReader::metadata`. In these cases the `ProbeResult` metadata queue will be empty.

I think an enum may be a bit much for this, but I could see wrapping the queue in an `Option<>`. Better documentation could also work.

BlackHoleFox:
Thank you immensely for the primer, that clears a lot of the questions up I had looking at the code. Knowing that, I agree that the enum is overkill and doesn't make much sense given what *could* be there.  I just pushed another revision (which I'll squash if this is approved), let me know what you think. 

Total sidenote but;
> For example, the ID3v2 tags on MP3 files are actually encoded in such a way that a MP3 decoder would ignore it as invalid data. A compatible decoder however will notice the tags and read them.

That seems cursed, but good to know :D

symphonia-core/src/probe.rs
```diff
@@ -161,8 +161,9 @@ pub struct ProbeResult {
     /// An instance of a `FormatReader` for the probed format
     pub format: Box<dyn FormatReader>,
     /// A queue of `Metadata` revisions read during the probe operation before the instantiation of
-    /// the `FormatReader`.
-    pub metadata: MetadataQueue
+    /// the `FormatReader`. If the container format itself contained the metadata, this will be be `None`. 
+    /// Metadata should then be obtained by calling `.metadata()` on `format`.
```
pdeljanov, line 165: This comment isn't quite accurate. It is possible for both the container format to contain metadata, and for the file/stream to have additional metadata "tacked on" to it outside of the container. In that case, the additional "tacked on" metadata is returned in this queue, while the metadata in the container is returned by `FormatReader::metadata()`.

pdeljanov (commented):
This looks good other than that one documentation issue.

Please also consider adding yourself to the CONTRIBUTORS file, especially since you have another PR that will most likely result in a code change.

pdeljanov:
> Total sidenote but;
> 
> > For example, the ID3v2 tags on MP3 files are actually encoded in such a way that a MP3 decoder would ignore it as invalid data. A compatible decoder however will notice the tags and read them.
> 
> That seems cursed, but good to know :D

Extremely cursed!

symphonia-core/src/probe.rs
BlackHoleFox: Sorry for the long delay, work got busy and some stuff came up I was putting off in person. I updated the doc comment to reflect this so please take another look when you get the chance.

pdeljanov:
Thanks, this was a good improvement to the readability of the API!

PR #35, opened 2021-06-30
Add initial fuzzing harness

5225225:
This, as-is, will try and fuzz all the default-enabled targets using the format guessing. This is *probably* good enough to not need one fuzzer target per format?

### DIFF
--- a/symphonia/fuzz/Cargo.toml
+++ b/symphonia/fuzz/Cargo.toml
@@ -14,6 +14,7 @@ libfuzzer-sys = "0.4"
 
 [dependencies.symphonia]
 path = ".."
+features = ["mp3", "aac"]
 
 # Prevent this from interfering with workspaces
 [workspace]
--- a/symphonia/fuzz/Cargo.toml
+++ b/symphonia/fuzz/Cargo.toml
@@ -1,0 +1,26 @@
+
+[package]
+name = "symphonia-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+[dependencies.symphonia]
+path = ".."
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "decode_any"
+path = "fuzz_targets/decode_any.rs"
+test = false
+doc = false
--- a/symphonia-bundle-flac/src/frame.rs
+++ b/symphonia-bundle-flac/src/frame.rs
@@ -199,7 +199,7 @@ pub fn read_frame_header<B: ByteStream>(reader: &mut B, sync: u16) -> Result<Fra
     // Get expected CRC8 checksum from the header.
     let crc8_expected = reader_crc8.into_inner().read_u8()?;
 
-    if crc8_expected != crc8_computed {
+    if crc8_expected != crc8_computed && cfg!(not(fuzzing)) {
         return decode_error("computed frame header CRC does not match expected CRC");
     }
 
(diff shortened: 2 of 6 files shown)


Shnatsel:
For FLAC you'll probably also want to disable checksum checks, since fuzzer-generated data doesn't have valid checksums in it. 

[Example from another Rust FLAG decoder](https://github.com/ruuda/claxon/commit/a5a9d3360527393d37cfe776eab5c8cb5927e5fc)

[Docs on conditional compilation during fuzzing](https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing)

And also [this article](https://shnatsel.medium.com/how-ive-found-vulnerability-in-a-popular-rust-crate-and-you-can-too-3db081a67fb) on how I fuzzed the other FLAC decoder might be relevant.

symphonia/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+[dependencies.symphonia]
+path = ".."
+
```
Shnatsel, line 17:
```suggestion
features = ["mp3", "aac"]
```

Shnatsel: Actually, that finds an underflow in `mp3` immediately, without even feeding it a starting corpus.

Shnatsel:
This finds issues almost immediately, and even without being fed a starting corpus, so you're on the right track :+1: 

FYI, `cargo fuzz` can be invoked with `--release` flag to ignore overflow and debug assertions. That should let you get to bugs with greater impact first. Optimizations are enabled even without the `--release` flag.

5225225:
The fuzzer finds an OOM (if you run it in `--release` to skip the overflow check), so the crate might need an API to specify a memory allocation limit. As in, "if you try to use more than *x*MB of memory to parse this, bail out". Then default that limit to something reasonable, high enough to not be an issue on basically all source files, low enough to not be vulnerable to a denial of service despite the limit.

That can be done after this goes in, but it's something that basically every parser needs to handle. (and hardly any of them *do*)

symphonia/fuzz/Cargo.toml
5225225: Yeah, I've found that for fuzzing, shallow bugs generally don't *need* starting corpuses, the fuzzer is pretty good at working out the format by itself.

pdeljanov:
Thanks for this, I haven't had time yet, but I'll try to test this out on the weekend. I scared to know how many bugs this'll find :stuck_out_tongue_closed_eyes: .

Shnatsel (approved):

pdeljanov:
Thanks to the both of you taking the initiative on this.

This is neat and I indeed do run into a panic in the MP3 decoder almost instantly. Will have to look into that one!

An issue I see going forward is that FLAC is going to be difficult to properly fuzz because I rely on the CRCs to packetize the stream (it's basically impossible otherwise without decoding). I don't think Claxon has to deal with that issue since I don't think it would go through the extra work to packetize the stream for no reason?

Shnatsel:
Do I understand correctly that packetization uses the CRC of the decoded data, so you can't just calculate it over the compressed stream without decoding it?

pdeljanov:
Not quite.

Basically there's two CRCs in FLAC. A CRC8 of the header at the end of the header, then a CRC16 of the header + payload data at the end of packet.

The FlacReader scans the stream for what looks like FLAC packet headers and then stores the potential header and whatever data it consumes before the next valid FLAC packet header is found. However, audio data can look like a valid FLAC packet header, so this method may be tricked into cutting off a packet too early. I therefore call this a presumptive packet a "fragment". So what FlacReader does is calculate the CRC16 of the fragment and checks if the last two bytes of the fragment matches the calculated CRC16. If it does, then we assume the fragment is infact a complete FLAC packet. If it does not match then we read a second fragment and repeat, except this time we calculate the CRC16 over the two fragments. This can repeat a few times until a  complete FLAC packet is found, or some heuristics kick in to "give up" on the initial fragment and assume the stream was corrupted and move on.

All this needs to be done because FLAC packets don't state a payload size...

Shnatsel:
Can we decide where we want the packet to end, then, and fabricate the appropriate CRC16?

pdeljanov:
Hmm, yeah, if you can generate random data, then calculate and place the correct CRC8 and CRC16 values into that buffer it should bypass the issue.

There's also value in stressing the FlacReader with extremely malformed streams and exercise those heuristics.

PR #36, opened 2021-06-30
Replace deprecated constant in symphonia-play

Herohtar:
Minor change that replaces the deprecated PulseAudio format constant with the new one.


pdeljanov:
Thanks!

PR #37, opened 2021-07-06
Remove unsafe code in chan_pair_mut

Herohtar:
Reduces unsafe code by 50%! :upside_down_face: 

The unsafe code block in `chan_pair_mut` can be removed by using `split_at_mut`. In order to preserve the original behavior of the method, a conditional needs to be used to split using the correct index depending on the order of the channels (0,1 vs 1,0). However, the code could be reduced if the method were changed to require the channels be requested in ascending order (first < second). The method is always called in that manner, but I wasn't sure if the ability to get channels in other combinations might be necessary in the future.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -15,6 +15,7 @@ Kostya Shishkov <kostya.shiskov@gmail.com>
 # Please keep this section sorted in ascending order.
 
 djugei [https://github.com/djugei]
+Herohtar [https://github.com/herohtar]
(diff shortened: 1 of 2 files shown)


pdeljanov:
:laughing: This is really good!

I think preserving the behaviour was a good idea. While it's not needed right now, some of the more modern codecs (AAC with >2 channels, Opus) can couple, and thus decode, arbitrary channels together. By being agnostic to the order in this function the logic will be simplified in those decoders.

Thanks!

Issue #38, opened 2021-07-06
Incorrect value used to construct SampleWriter


Herohtar:
`SampleWriter::from_buf` uses an unsafe code block to create a mutable slice, [with `buf.capacity()` as the length](https://github.com/pdeljanov/Symphonia/blob/81e9d5c31d3cbfd0ee51ebd969d264794efa875b/symphonia-core/src/audio.rs#L960); however, it seems the intention is for the referenced slice to be of length `n_samples`. It appears that the method is always called with `n_samples` equal to `buf.capacity()`, but it could cause unintended behavior if the code is ever called with a different value.

pdeljanov:
This is indeed an issue. Good catch!

The purpose of `SampleWriter` and the `WriteSample` trait was to write `Sample`s from an `AudioBuffer` into a `RawAudioBuffer` as raw bytes. In C/C++, it'd be possible to just allocate a buffer and cast the buffer pointer to the actual sample data type (u24 being the exception) and then copy the samples, or even just use a memcpy. This code was trying to emulate that implementation in Rust for "performance." However, I now think a safe implementation would be just as fast and cleaner.

Therefore, I actually think this part of the code needs to be rewritten entirely. Once this is rewritten safely then the library will be 100% safe Rust.

Is this something you would be interested in taking on?

Herohtar:
I was actually initially hoping to be able to replace both instances of unsafe code, which is how I discovered that issue. My existing Rust knowledge wasn't sufficient for me to determine a solution right away and my brief research into it only resulted in more unsafe methods, but I'd definitely be interested in continuing to look into rewriting it.

PR #39, opened 2021-07-08
Remove unsafe from SampleWriter

BlackHoleFox:
Removes the last use of `unsafe` from symphonia.

Obligatory disclaimer that while this compiles fine and `cargo test` passes, I'm not sure how to actually test these changes on actual audio since `symphonia-play` doesn't like my computer's audio devices for some reason, so im blind to if this actually broke something.

Closes https://github.com/pdeljanov/Symphonia/issues/38


BlackHoleFox:
Marked as draft since I completely missed that `as_bytes` is part of the public API. Ill need to think a bit more there, I suppose.

pdeljanov:
Hmmm, I'm not opposed to using an external crate such as `bytemuck` if it's hidden and not part of the public interface. But, it looks like we need to mark `Sample` as `bytemuck::Pod`? I'm not particularly keen on this.

`RawAudioBuffer` is basically a reverse `symphonia_core::io::ByteStream` so it should be possible to do with safe code. Performance would be a question mark though. If it came down to it, we could also limit the unsafe code to this module only...

BlackHoleFox:
> . But, it looks like we need to mark Sample as bytemuck::Pod? I'm not particularly keen on this.

Didn't run into this myself. I only had to make `Sample::StreamType` implement `Pod` in order to get the functionality of `as_bytes`. 

> RawAudioBuffer is basically a reverse symphonia_core::io::ByteStream so it should be possible to do with safe code. Performance would be a question mark though. If it came down to it, we could also limit the unsafe code to this module only...

Its trivially easy to do with safe code is you bring allocations to the table, but that would tank performance when you're just trying to hand a view of bytes off to some audio player (like `pulseaudio`.)

I created a Linux VM to try this out in and as far as I can tell, it played the stream though for the expected ~30 seconds of the clip, although the VM's sound was very broken so I couldn't hear anything anyway.

pdeljanov:
> > . But, it looks like we need to mark Sample as bytemuck::Pod? I'm not particularly keen on this.
> 
> Didn't run into this myself. I only had to make `Sample::StreamType` implement `Pod` in order to get the functionality of `as_bytes`.
> 

My bad, I mean't `Sample::StreamType`, which is also technically the public interface. Though, I haven't tested the feasibility of an external user implementing a custom `Sample` type.

> > RawAudioBuffer is basically a reverse symphonia_core::io::ByteStream so it should be possible to do with safe code. Performance would be a question mark though. If it came down to it, we could also limit the unsafe code to this module only...
> 
> Its trivially easy to do with safe code is you bring allocations to the table, but that would tank performance when you're just trying to hand a view of bytes off to some audio player (like `pulseaudio`.)

Yeah, we don't want to allocate during decode  :laughing:. Worst-case would just be using `*::to_ne_bytes` on the `Sample` followed by `copy_from_slice` into the underlying `Vec<u8>`.

> I created a Linux VM to try this out in and as far as I can tell, it played the stream though for the expected ~30 seconds of the clip, although the VM's sound was very broken so I couldn't hear anything anyway.

Hmm, I'm assuming you either use a Mac or Windows PC because you didn't notice the breakage in `RawAudioBuffer` which is what the PulseAudio output uses. What's the problem exactly? I'd think `symphonia-play` should "just work" since it uses `cpal` on those OSs.

BlackHoleFox:
I use Windows day-to-day. When trying out `symphonia-play` it prints all the metadata but dies hard trying to actually play it:
```
ERROR symphonia_play::output::cpal > audio output stream open error: The requested stream configuration is not supported by the device.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OpenStreamError', symphonia-play\src\main.rs:248:74
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

BlackHoleFox:
> My bad, I mean't Sample::StreamType, which is also technically the public interface. Though, I haven't tested the feasibility of an external user implementing a custom Sample type.

ahh, that makes more sense now. Not an audio expert by any means, but there probably can't be that many sample types out there right? You (might) even be able to get away with sealing the `Sample` trait if theres likely not any reason for an external crates to implement it. I do agree that exposing third-party dependencies isn't great though if thats not reasonable :rip. On the flipside, `Pod` by default is implemented for most types that would really make sense to use in terms of samples (numbers and arrays)

> Yeah, we don't want to allocate during decode 😆. Worst-case would just be using *::to_ne_bytes on the Sample followed by copy_from_slice into the underlying Vec<u8>.

Thats pretty much what I was thinking, since you'd need an `O(N)` second allocation to push all the bytes into to get a contiguous view of bytes and not something like an iterator.

pdeljanov:
> I use Windows day-to-day. When trying out `symphonia-play` it prints all the metadata but dies hard trying to actually play it:
> 
> ```
> ERROR symphonia_play::output::cpal > audio output stream open error: The requested stream configuration is not supported by the device.
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OpenStreamError', symphonia-play\src\main.rs:248:74
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Hmmm, the `CpalAudioOutput` only outputs in floating-point, or signed/unsigned 16-bit audio so those should be well supported on anything above Windows XP... I guess the sample rate could be an issue, but that seems odd. Are you running your audio output in exclusive mode? 

> > Yeah, we don't want to allocate during decode laughing. Worst-case would just be using *::to_ne_bytes on the Sample followed by copy_from_slice into the underlying Vec.
> 
> Thats pretty much what I was thinking, since you'd need an `O(N)` second allocation to push all the bytes into to get a contiguous view of bytes and not something like an iterator.

Well, there's always going to be an `O(N)` copy regardless since Symphonia stores the audio in a planar-style (all the samples of channel N, before all the samples of channel N+1, and so on) in `AudioBuffer`, whereas most outputs need it in interleaved-style (the first sample for each channel are stored consecutively, before the second samples for each channel, and so on). Even if the output is also planar, it only becomes copyless if the `AudioBuffer` is 100% full, otherwise there will be a gap.

BlackHoleFox:
> 
> 
> > I use Windows day-to-day. When trying out `symphonia-play` it prints all the metadata but dies hard trying to actually play it:
> > ```
> > ERROR symphonia_play::output::cpal > audio output stream open error: The requested stream configuration is not supported by the device.
> > thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OpenStreamError', symphonia-play\src\main.rs:248:74
> > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> > ```
> 
> Hmmm, the `CpalAudioOutput` only outputs in floating-point, or signed/unsigned 16-bit audio so those should be well supported on anything above Windows XP... I guess the sample rate could be an issue, but that seems odd. Are you running your audio output in exclusive mode?

It turned out it didn't like my primary audio device. Switching over to a DAC fixed the issue. Sadly that doesn't help me any with the `RawSampleBuffer` issue since it doesn't use it though D: Can you describe what broke with it? I wonder if its an issue of what `as_bytes()` is returning or if I messed up an offset refactoring inside it. My bet is on one of the `* mem::size_of::<S::StreamType>();` calls I removed doing it.

> 
> > > Yeah, we don't want to allocate during decode laughing. Worst-case would just be using *::to_ne_bytes on the Sample followed by copy_from_slice into the underlying Vec.
> > 
> > 
> > Thats pretty much what I was thinking, since you'd need an `O(N)` second allocation to push all the bytes into to get a contiguous view of bytes and not something like an iterator.
> 
> Well, there's always going to be an `O(N)` copy regardless since Symphonia stores the audio in a planar-style (all the samples of channel N, before all the samples of channel N+1, and so on) in `AudioBuffer`, whereas most outputs need it in interleaved-style (the first sample for each channel are stored consecutively, before the second samples for each channel, and so on). Even if the output is also planar, it only becomes copyless if the `AudioBuffer` is 100% full, otherwise there will be a gap.

Thats good to know for the future. I was thinking about the simpler, local effect, to `RawSampleBuffer` of avoiding adding an `O(1)` allocation for the decoding process since I figured the rest of the crate wasn't entirely zero-copy since its usually like this with parsers,
[ ] fast
[ ] safe
[ ] efficient 
 (pick 1 or 2)

pdeljanov:
> > > I use Windows day-to-day. When trying out `symphonia-play` it prints all the metadata but dies hard trying to actually play it:
> > > ```
> > > ERROR symphonia_play::output::cpal > audio output stream open error: The requested stream configuration is not supported by the device.
> > > thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OpenStreamError', symphonia-play\src\main.rs:248:74
> > > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> > > ```
> > 
> > 
> > Hmmm, the `CpalAudioOutput` only outputs in floating-point, or signed/unsigned 16-bit audio so those should be well supported on anything above Windows XP... I guess the sample rate could be an issue, but that seems odd. Are you running your audio output in exclusive mode?
> 
> It turned out it didn't like my primary audio device. Switching over to a DAC fixed the issue. Sadly that doesn't help me any with the `RawSampleBuffer` issue since it doesn't use it though D: Can you describe what broke with it? I wonder if its an issue of what `as_bytes()` is returning or if I messed up an offset refactoring inside it. My bet is on one of the `* mem::size_of::<S::StreamType>();` calls I removed doing it.
> 

Can't say I know what's up with your primary audio device vs. DAC, but in terms of your code change: `RawSampleBuffer::capacity()` should just return `self.buf.len()`. The division is no longer necessary because the underlying vector doesn't hold bytes anymore. Can confirm this works on Linux properly now.

That being said, still not keen on introducing `bytemuck` like this.

> > > > Yeah, we don't want to allocate during decode laughing. Worst-case would just be using *::to_ne_bytes on the Sample followed by copy_from_slice into the underlying Vec.
> > > 
> > > 
> > > Thats pretty much what I was thinking, since you'd need an `O(N)` second allocation to push all the bytes into to get a contiguous view of bytes and not something like an iterator.
> > 
> > 
> > Well, there's always going to be an `O(N)` copy regardless since Symphonia stores the audio in a planar-style (all the samples of channel N, before all the samples of channel N+1, and so on) in `AudioBuffer`, whereas most outputs need it in interleaved-style (the first sample for each channel are stored consecutively, before the second samples for each channel, and so on). Even if the output is also planar, it only becomes copyless if the `AudioBuffer` is 100% full, otherwise there will be a gap.
> 
> Thats good to know for the future. I was thinking about the simpler, local effect, to `RawSampleBuffer` of avoiding adding an `O(1)` allocation for the decoding process since I figured the rest of the crate wasn't entirely zero-copy since its usually like this with parsers,
> [ ] fast
> [ ] safe
> [ ] efficient
> (pick 1 or 2)

The decoders are all zero-allocation after init, but the format readers are not (always need to allocate for a `Packet`, but that's amortized over many audio samples).

Generally though, `SampleBuffer`/`RawSampleBuffer` should be reused except when the `SignalSpec` changes. So they should be allocated at init. Mind you, nothing is stopping developers from just using the `AudioBuffer` directly and grabbing samples from there to feed the audio output device. The sample buffers just make taking audio data out of Symphonia easier in the general use-case.

thomcc:
(Hi, thanks for Symphonia)

The code as-is is actually unsound currently. RawSampleBuffer hands out memory from a `Vec<u8>`, which you then convert to (more or less) an `&mut [<impl Sample>::StreamType]`. 

`Vec<T>` only requests memory with alignment of `align_of::<T>()` from the global allocator. In the case of `u8`, this is align of 1<sup>*</sup>. Many of the stream types have higher alignments. The only ones that don't are u8, i8, u24, i24 (u24 and i24 have an array of 3 bytes as a stream type, and so they have an alignment of 1). This is dodgy, and is relying on things that aren't guaranteed to avoid UB, unfortunately.

---

Anyway, I think even using bytemuck here will not be a great option, it's an improvement, since it detects the case where things go wrong, but it... would panic in that case.

I'd honestly suggest just always reading bytes out of the stream and having a decode step (whcih would generally be a from_le_bytes or from_be_bytes depending on the endianness of the format)...

E.g. Maybe forgo the stream type all together, and just assume you'll be reading out of the bytes, which avoids assuming anything about alignment. I think that will save some future headaches on big endian systems too (or on LE systems with BE PCM data), but maybe this is too painful (or you have other plans on how to handle that)

---

\* Note: The common system allocators (which are used by default) will all hand out memory with at least alignment of `alignof(maxalign_t)` as they're written for C, and this is required by C. This is usually 8 or 16, but always guaranteed to be enough for any of the primitive C types on the system, which means it's also going to be enough for any of the current stream types.

However, other allocators (e.g. from when `#[global_allocator]` is overridden) are totally free to hand out allocations with only the alignment requested. This (along with the fact that on x86 unaligned loads tend not to fault) is probably why the code works fine currently.

pdeljanov:
Thanks @thomcc for pointing out that issue and the good explanation.

I think a decode/encode style solution where we use `from/to_ne_bytes` would work, though, I think it would need to be benchmarked before committing to it. I can't really think of a case where on a LE machine we'd need to export BE PCM audio, or vice-versa. Atleast.. not with the use case I had in mind for `RawAudioBuffer` which is to get PCM audio samples out of Symphonia to feed the OS's audio API.

If we stick with an unsafe solution, then I think the problem you described could be fixed by overallocating the `Vec<u8>` by 16 bytes and then manually aligning the pointer to the first `StreamType` of the slice. Does this sound reasonable?

thomcc:
> If we stick with an unsafe solution, then I think the problem you described could be fixed by overallocating the Vec<u8> by 16 bytes and then manually aligning the pointer to the first StreamType of the slice. Does this sound reasonable?

Yes, that works. Using `v.as_slice().align_to::<T>().1` (https://doc.rust-lang.org/std/primitive.slice.html#method.align_to) is an okay option for this, and bytemuck has an [equivalent](https://docs.rs/bytemuck/1.7.0/bytemuck/fn.pod_align_to.html).

Anyway, I realized you can use bytemuck for this without it being in your public API so long as you're willing to [seal](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) the stream type:

```rs
mod sealed {
    pub trait Sealed: bytemuck::Pod {}
}

pub trait StreamType: Copy + Sized + sealed::Sealed + 'static {}

// Impl for all the values allowed as Sample::StreamType...
impl sealed::Sealed for u8 {}
impl StreamType for u8 {}
impl sealed::Sealed for i8 {}
impl StreamType for i8 {}
impl sealed::Sealed for u16 {}
impl StreamType for u16 {}
impl sealed::Sealed for i16 {}
impl StreamType for i16 {}
impl sealed::Sealed for u32 {}
impl StreamType for u32 {}
impl sealed::Sealed for i32 {}
impl StreamType for i32 {}
impl sealed::Sealed for f32 {}
impl StreamType for f32 {}
impl sealed::Sealed for f64 {}
impl StreamType for f64 {}
impl sealed::Sealed for [u8; 3] {}
impl StreamType for [u8; 3] {}

pub trait Sample {
    type StreamType: StreamType;
    // ..
}
```

This doesn't end up with bytemuck in your public API anywhere, or in your documentation, but you can use it internally.

Then you can replace your `slice::from_raw_parts_mut(bytes.as_mut_ptr() as *mut S::StreamType, buf.capacity())` with a `bytemuck::pod_align_to_mut(bytes).1`, which is safe.

This would also actually allow custom sample types, so long as they use one of the supported stream types. IDK how useful this is, though.

---

P.S. Note that there's no way to get a Vec or `Box<[u8]>` with an over-aligned backing buffer, which would be a less obtrusive fix I guess. The problem is they pass the alignment in to dealloc, and it needs to match — there are allocators that keep alignments in different buckets. You aren't suggesting this, but it's something that might be tempting and is unintuitively disallowed

pdeljanov:
After mulling over it, it's probably unlikely that custom `Sample` formats will ever be implemented. Mainly because the `AudioBufferRef` enum returned by `decode` can't be extended externally and thus there's no ability to return an `AudioBuffer` of a custom sample format. The only other use-case then would be if the host OS wants some weird sample format (i36?, i48?, u56?) and the user wants to use a `SampleBuffer` to do the conversion. Again, pretty tenuous, and arguably we could just implement the extra formats now and call it a day.

In that case, this sealed approach seems like a good option. I'll try updating this PR with the changes. I think I'll also remove the `StreamType` from the `Sample` trait and instead add it to the `WriteSample` trait (probably rename it to `RawType` too), then we can keep all of the machinery for this within the `audio` module.

pdeljanov:
Made those changes and then some. 

I removed the `SampleWriter`, and rewrote `RawSampleBuffer`'s `copy_` functions to use iterators instead. Theoretically it should be faster.

@thomcc, you mentioned this:

> P.S. Note that there's no way to get a Vec or `Box<[u8]>` with an over-aligned backing buffer, which would be a less obtrusive fix I guess. The problem is they pass the alignment in to dealloc, and it needs to match — there are allocators that keep alignments in different buckets. You aren't suggesting this, but it's something that might be tempting and is unintuitively disallowed

Since in the latest PR revision we use `Vec<S::RawType>` to store the raw samples, we will get a reference to `&[S::RawType]` and then cast it to `&[u8]` using bytemuck. This should be safe because `u8` effectively doesn't care about alignment and we aren't freeing the `&[u8]`. But I'm getting the impression from what you wrote that this is may be a bad assumption. Am I missing something?

thomcc:
Nope, that’s fine. What would have been bad is using std::alloc::alloc to allocate a ptr with the desired alignment, then using Vec::from_raw_parts to put it in a `Vec<u8>`.

pdeljanov:
Thanks @BlackHoleFox and @thomcc. Great discussion and suggestions, I've merged this PR.

BlackHoleFox:
I’m glad we got the best resolution out of this, thanks for the great back-and-forth.

Shnatsel:
Since there were some alignment issues in the old code, could you publish a new release incorporating these changes?

pdeljanov:
Hey @Shnatsel, I'll be publishing v0.4.0 in about 2-3 weeks. Lots of changes & a new Vorbis decoder in this release so just cleaning things up and dogfooding it a bit before releasing.

Shnatsel:
I suppose v0.3.1 with this fix would be useful, but I don't really have a horse in this race, so up to you.

thomcc:
This is a breaking change so it can't be part of 0.3.1.

That said, I might be misremembering (it was a while ago), but IIRC the alignment issue were in proposed APIs and not APIs that exist in 0.3.0, aside from one case where a weird user implementation of the `Sample` trait could allow a safe API to expose a slice with the wrong alignment.

In practice this is probably pretty unlikely, since pretty much any sample types you could want is supported (it's obviously important to fix, but IDK if anybody's code actually has this problem).

Maybe there's an issue I'm forgetting though?

pdeljanov:
Indeed, this change did modify the public interfaces, albeit, ones that would've been unlikely to be used by anyone directly. However, to be true to semver we need to bump to v0.4.0. I decided to roll multiple large breaking interface changes into this release which is why this fix hasn't been published yet.

Shnatsel:
In that case, the plan to cut v0.4 in a month or so sounds good. Thanks!

PR #42, opened 2021-07-11
Fix symphonia-play Windows build

Herohtar:
Fixes a type that didn't get changed when `WriteSample` was replaced with `RawSample`

### DIFF
--- a/symphonia-play/src/output.rs
+++ b/symphonia-play/src/output.rs
@@ -139,7 +139,7 @@ mod cpal {
     trait AudioOutputSample :
         cpal::Sample +
         ConvertibleSample +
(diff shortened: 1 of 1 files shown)


BlackHoleFox:
Funnily enough I'm currently working on improving CI as a PR, including running it in Windows. Hopefully that would prevent breakages like this.

pdeljanov:
Big yikes with this one. Yeah, `symphonia-play` doesn't get built by the CI and I mainly develop on Linux so it's too easy to miss these things..

BlackHoleFox (approved):
Confirmed this fixes `cargo check --all` on Windows :)

PR #44, opened 2021-07-11
improved(ci): Overhaul and improve GitHub Actions

BlackHoleFox:
Based off [`rusqlite's` setup](https://github.com/rusqlite/rusqlite/blob/master/.github/workflows/main.yml) and @thomcc's [blog](https://shift.click/blog/github-actions-rust/) for the bases, plus what I've seen done in other projects.

Ideally this should stop things like https://github.com/pdeljanov/Symphonia/pull/42 from happening in the future, or other subtle bugs if symphonia's functionality expands for some reason.

Ideally the `check` steps would be replaced by `clippy` instead once the workspace has has a bunch of the current warnings cleaned up since it generally helps keep the codebase clean and consistent. Rustfmt too, for similar reasons.


BlackHoleFox:
Not sure why the Windows check is failing when the `test` passes, nor is GitHub giving me much info:
image

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+  pull_request: {}
+
+jobs:
+  check:
+    name: Check ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, "windows"]
```
pdeljanov, line 16: I think this may have to be `windows-latest`?

BlackHoleFox: :facedesk: Thank you, completely missed that.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all
+        
+    - name: Check all features
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --all-features --exclude symphonia-play
```
pdeljanov, line 39: If `symphonia-play` is excluded, would this still actually catch the issue seen earlier today?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all --doc
+
+      - name: Test all features
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all --all-features --doc --exclude symphonia-play
```
pdeljanov, line 69: Ditto above

.github/workflows/ci.yml
BlackHoleFox: The default features test should catch it (and `cargo check —all` caught it locally). I don’t think `symphonia-play` needs to be checked/tested with `—all-features` since it has none of its own features defined.

pdeljanov: Sounds good!

PR #48, opened 2021-07-11
Run all the tests in CI

BlackHoleFox:
This follows up https://github.com/pdeljanov/Symphonia/pull/44 to fix a small issue I noticed while looking at addressing https://github.com/pdeljanov/Symphonia/issues/47.

Between a bit of misreading of the `cargo test` documentation and a lot of churn before I put my PR up, I accidentally used `--doc` for the only `--all-features` test, meaning the actual tests won't run.

Sorry for the PR churn :rip:

PR #49, opened 2021-07-12
Run tests on PowerPC for big-endian coverage

### LINKED ISSUE #47: [Task] Test on a big-endian virtual machine
Symphonia should work properly on both big and little-endian devices. x86/x86_64 and ARM devices are all little-endian, but Power (and MIPS?) architecture devices can run in big-endian mode.

Therefore, let's explore using a virtual machine (QEMU?) to test if Symphonia works on big-endian devices and fix any issues found. If at all feasible, could this also be integrated into CI?

BlackHoleFox:
The Cargo actions-rs package has a [nice hook-in](https://github.com/actions-rs/cargo#cross-compilation) for `cross`, which should let us run the tests on a big-endian architecture, in this case 32-bit PowerPC, pretty painlessly in CI.

This is currently based off https://github.com/pdeljanov/Symphonia/pull/48.

Closes #47

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,12 +39,16 @@ jobs:
         args: --all --all-features --exclude symphonia-play
 
   test:
-    name: Test ${{ matrix.os }}
+    name: Test ${{ matrix.config.target }} on ${{ matrix.config.os }}
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.config.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, "windows-latest"]
+        config:
+          - { os: ubuntu-latest, target: x86_64-unknown-linux-gnu, use-cross: false }
+          - { os: macos-latest, target: x86_64-apple-darwin, use-cross: false }
+          - { os: windows-latest, target: x86_64-pc-windows-msvc, use-cross: false }
+          - { os: ubuntu-latest, target: powerpc-unknown-linux-gnu, use-cross: true }
 
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


pdeljanov:
This is a lot less painful than I thought it would be!

Some bad news though, while the CI integration seems to work, I don't think this is actually testing anything meaningful yet in the context of endianness. That's because it doesn't look like any of the `MediaSourceStream` tests actually test the `read_le/be_*` functions. The current suite of tests just use the `read_byte` and `read_double/triple/quad_bytes` functions which read bytes in the order they appear in the stream.

Not sure if you want to tackle those with this PR though. You don't have to, but we can't close #47 yet.

BlackHoleFox:
@pdeljanov sorry for my slowness, I have a bad habit of overloading myself.

Would you mind expanding on what you'd like to see for test cases of the `be/le` methods? I'm not informed enough to really know what constitute a full test. I added some stubs, but they're not great at all and test the bare minimum. Do you want something like `verify_mss_read` but that uses some of the `_be_` and `_le` methods in a kinda similar way?

BlackHoleFox:
I have no clue if its because I messed up my tests or not, but the MIPS test job is [failing with different values](https://github.com/pdeljanov/Symphonia/runs/3120255852?check_suite_focus=true).

thomcc:
MIPS has a particularly idiosyncratic implementation of IEEE 754. I believe it's not even compliant with IEEE 754-2008 (they chose backwards compatibility over standards compliance, after encoding some things differently than everybody else — admittedly this is mostly about how NaNs work). You're probably better off using a PPC target as a testbed.

That said, this probably is actually a bug in the code.

pdeljanov:
Hey @BlackHoleFox, no worries. Don't feel pressured to spend a lot of time on this project. Things take the time they take.

I just wanted to point out that there were no pre-existing tests for the BE/LE read functions so a passing check would be misleading.

The tests you added are basically what I was looking for. As long as the bit patterns for the values being tested aren't symmetric and thus the same on BE and LE machines we're good.

The issue you ran into stumped me for a bit, but I think I figured it out. The `generate_random_bytes` function is a random number generator and part of the algorithm involves converting a `u32` to `[u8; 4]` which is done via `to_ne_bytes`. And therein lies the problem, the bytes being generated will be different depending on the system's endianness. I'll mark this as a comment in the review. I think we can just change `to_ne_bytes` to `to_le_bytes` so your tests don't need to be updated.

As an interesting side note, the other tests that use `generate_random_bytes` didn't break because they compare the data read from the `MediaSourceStream` directly with the actual buffer that was generated. Thus, the actual data in that buffer didn't really matter at all!

BlackHoleFox:
That looks to have fixed it, good catch :D It looks like CI is working happily now.

> As an interesting side note, the other tests that use generate_random_bytes didn't break 

Fascinating, and that would explain why only mine ended up going sideways.

pdeljanov:
Excellent! Thanks for this!

PR #53, opened 2021-09-12
core: Allow access to `Decoder`'s last frame.

FelixMcFelix:
Adds `Decoder::last_decoded` to expose read access to internal buffers after a call to `decode`. This is useful when dealing with multiple tracks/files to be mixed into a single output buffer, particularly if this output buffer is smaller its input frames (e.g., 20ms vs. 26ms MP3 frames). `AudioBufferRef`s can't be safely stored alongside their parent decoders, so this ensures that any leftover data doesn't need to be preemptively copied out.

Currently, this exposes `buf` in all decoder implementations if the frame count is nonzero, and expects the user not to call `last_decoded` if the previous frame caused a decoding error.


pdeljanov:
Hi @FelixMcFelix,

Thanks for this, and apologies for the delay. This is a good idea, but I have reservations about the error case.

Currently, most of the decoders are using the output buffer as a scratch buffer to save on memory and extra copies. Depending on where the decoding failed, if the buffer were played back, you may get noise that far exceeds the normal range (e.g., for floating point buffers, samples that are greater-than +/-1.0).

In the current design, this buffer isn't accessible unless decoding is successful. However, if we make it accessible via. `Decoder::last_decoded` we open an avenue to access invalid data. I feel this would become a fairly big API footgun that could damage equipment and even a listener's hearing.

I see a way forward if we change the definition of `Decoded::last_decoded` to return the last *successfully* decoded buffer. This means we'll have to modify the affected decoders to allocate their own scratch buffer, and only copy the result into the final buffer when successful.

In terms of performance this should be fine since the final step of most decoders is to clamp or scale the output. Instead of being in-place, we'll just write the output into the non-scratch buffer. The only cost will be memory, but it seems like in practice the application will have to make a copy anyways.

Thoughts?

FelixMcFelix:
Hi @pdeljanov,

Thanks, it's no issue. I haven't had too much time lately to work on much open source so it's been out of mind for a while. The main issue you have makes sense, in that making it easier for users to have access to invalid data wasn't really my intent!

The *last successfully decoded buffer* is a better assumption, but I think doubling memory consumption in the general case is a little hefty. This extra buffer is the only way to guarantee that such a buffer view always exists after an error, though. Could we weaken the requirement to "returns the last frame if the last call to `Decoder::decode` succeeded"? The simplest possible way to do this is to convert any existing `decode` implementations into `decode` and `decode_inner` (and then having any `Err`s invalidate the buffer), but this is probably ugly and repetitive.

I think this still avoids the footgun from the library user's perspective at lower memory cost, but I think both these approaches push some extra complexity (or at least busy-work) onto the implementer. This could be moved into another trait so that `Decoder`s don't need to necessarily provide this, but that makes it a bit less convenient to use too.

pdeljanov:
> Thanks, it's no issue. I haven't had too much time lately to work on much open source so it's been out of mind for a while. The main issue you have makes sense, in that making it easier for users to have access to invalid data wasn't really my intent!

Ahh, the open source struggle. :sweat_smile: 

> Could we weaken the requirement to "returns the last frame if the last call to `Decoder::decode` succeeded"? The simplest possible way to do this is to convert any existing `decode` implementations into `decode` and `decode_inner` (and then having any `Err`s invalidate the buffer), but this is probably ugly and repetitive.

This is something I thought of too, and it probably makes more sense to a user as well: If `Decoder::decode` failed, why would `Decoder::last_decoded` return something at all? However, it does mean that the every decoder will need to track an extra bit of state recording the result of the last decode call.

There is one other option though. How about returning an empty buffer on error. If `Decoder::decode_inner` fails then the audio buffer can simply be cleared in `Decoder::decode`. This would remove the need to maintain that extra state and may actually make decode loops quite a bit cleaner (an error yields you a 0 sized buffer so you'd simply loop to decode the next packet which is what you'd want to do in the non-error case as well).

FelixMcFelix:
Thanks for the feedback. I've changed `last_decoded` to just give access to `buf` (which will be valid, or zero-length), and ported existing codecs to `decode` and `decode_inner` (with a buffer clear on error). The `Decoder` trait now specifies that clearing the inner buffer on `Err` is a requirement.

I'm a little unsure as to what you meant about simplifying `decode` loops, since I expect that would still need to return an error?

I've tested this on various supported codecs (MP3, Vorbis, ALAC, FLAC, a local libopus wrapper) in a really messy mixing chain (2.5ms slices -> optional sample convert -> optional FFT resample -> mix into 20ms 48kHz buffer, with various scratch buffers for boundaries of packets). Songs sound and mix as expected.

pdeljanov:
Hey @FelixMcFelix,

This looks really good. Thanks for making these changes.

> I'm a little unsure as to what you meant about simplifying `decode` loops, since I expect that would still need to return an error?

I probably explained it poorly, but I was thinking that for many applications it's useful to have audio chunked into fixed duration blocks. Without your change, this is difficult because we need to copy the decoded audio buffer since it may be 
 larger than any one chunk. With your change it's now possible to do something like below.

```rust
pub struct Decoder {
    reader: Box<dyn symphonia::core::formats::FormatReader>,
    decoder: Box<dyn symphonia::core::codecs::Decoder>,
    track_id: u32,
    cursor: Range<usize>,
}

impl Decoder {
    fn next_audio_chunk(&mut self) -> Result<AudioChunk> {
        // let chunk = AudioChunk::new();
        loop {
            if !self.cursor.is_empty() {
                let decoded = self.decoder.last_decoded();

                // Copy audio from the decoded audio buffer to `chunk`.
                // Update the cursor.
                // Break if `chunk` is now full.
            }

            let packet = match self.reader.next_packet()?;

            if packet.track_id() != self.track_id {
                continue;
            }

            match self.decoder.decode(&packet) {
                Ok(decoded) => {
                    self.cursor = 0..decoded.frames();
                }
                Err(symphonia::core::errors::Error::DecodeError(err)) => continue,
                Err(err) => return Err(err.into()),
            }
        }

        Ok(chunk)
    }
}
```

I also think this is really clean and elegant because there's no special handling for `DecodeError`s or next packet cases.

pdeljanov (approved):

PR #58, opened 2021-09-23
AAC missing channel fix

5225225:
This fixes an error when decoding an invalid aac file found by the fuzzer.

The panic was on channels, but I assume validation of the sample rate existing here is also needed.


pdeljanov:
Thanks for the fix, and apologies for the delay.

The `AacDecoder` needs a substantial refactoring and cleanup, but the validation checks look good to me.

PR #61, opened 2021-10-16
Added missing Clone & Debug derives to core::meta

Beinsezii:
Every struct now has at least Clone and Debug with the exception of
Metadata<'a>. There might be more missing derives elsewhere but so far
this is the only part of the lib I've used directly.

### DIFF
--- a/symphonia-core/src/meta.rs
+++ b/symphonia-core/src/meta.rs
@@ -23,7 +23,7 @@ use crate::io::MediaSourceStream;
 ///
 /// All limits can be defaulted to a reasonable value specific to the situation. These defaults will
 /// generally not break any normal streams.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub enum Limit {
     /// Do not impose any limit.
     None,
@@ -52,7 +52,7 @@ impl Default for Limit {
 }
 
 /// `MetadataOptions` is a common set of options that all metadata readers use.
-#[derive(Copy, Clone, Default)]
+#[derive(Copy, Clone, Debug, Default)]
 pub struct MetadataOptions {
     /// The maximum size limit in bytes that a tag may occupy in memory once decoded. Tags exceeding
     /// this limit will be skipped by the demuxer. Take note that tags in-memory are stored as UTF-8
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hi @Beinsezii,

Thank you for cleaning up these inconsistencies.

I only have reservations about adding `Clone` to `Metadata`, `MetadataBuilder`, and `MetadataLog`. These are rather heavy objects that should only be moved. In the case of `Metadata` (an intermediary object to access `MetadataRevisions`), `Clone` doesn't make sense in practice.

Is there some use-case I'm missing that requires these objects to be cloneable?

Thanks!

Beinsezii:
IMO, Clone doesn't really need a 'use case' as sometimes you might just want two of something. It's the end-programmer's responsibility to not clone something if they don't need a copy, and restricting it on a library-level is kinda weird.

Possible cases:
 * For me I was toying with the idea of sending a copy of the data over a local socket for IPC.
 * While prototyping sometimes it's easier to use .clone() initially while you figure out how it's going to look, then put in refs/arcs/whatever later where it makes sense.
 * You might literally duplicate a track. Duping the structured metadata means you don't have to re-read the binary data again.

>  In the case of `Metadata` (an intermediary object to access `MetadataRevisions`), `Clone` doesn't make sense in practice.

I only added Debug to Metadata<'a> as it's impossible to clone it.

pdeljanov:
> IMO, Clone doesn't really need a 'use case' as sometimes you might just want two of something. It's the end-programmer's responsibility to not clone something if they don't need a copy, and restricting it on a library-level is kinda weird.

Gotcha, thanks for the explanation.

Your PR has gotten me wondering if there is a best practice documented somewhere regarding the usage of derive. For example, should libraries always derive `Debug` on public structs or would that bloat the build too much? If you've seen any good reading or recommendations on the matter, please let me know.

> I only added Debug to Metadata<'a> as it's impossible to clone it.

My mistake!

Beinsezii:
`Debug` specifically has been [brought up](https://www.reddit.com/r/rust/comments/8677z1/why_does_rust_not_derivedebug_by_default/) a [couple times](https://www.reddit.com/r/rust/comments/7y7tz0/is_there_a_reason_rust_does_not_derive_fmtdebug/) and it generally seems like if you liberally use `"{:?}"` in your final binaries it could increase filesize a bit, and deriving a large order of structs or an extremely massive struct could hurt compile times.

My personal takeaway is unless you're extremely limited by resources working on/for anemic hardware, or your project is intended to be massive in scale, I just say F it and defensively `#derive` my mods' public structs. Granted I don't think any of the random schizo shit I make is going to be used on a Pi Zero anytime soon.

PR #79, opened 2021-12-05
enable mp3 feature by default

Be-ing:
MP3 codec quality is listed as "Excellent" in README.md.

### DIFF
--- a/symphonia/Cargo.toml
+++ b/symphonia/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["audio", "codec", "decoder", "multimedia", "media"]
 edition = "2018"
 
 [features]
-default = ["wav", "flac", "pcm", "ogg", "vorbis"]
+default = ["wav", "flac", "pcm", "mp3", "ogg", "vorbis"]
 aac = ["symphonia-codec-aac"]
 alac = ["symphonia-codec-alac"]
 flac = ["symphonia-bundle-flac"]


pdeljanov:
The default set of codecs and formats is not based on the maturity of the feature. Only patent-free/royalty-free/open source codecs and formats are enabled by default.

I would accept an `all` feature that enables all codecs & formats. This could be useful. It could even be more granular with sub-categories such as `all-codecs`, and `all-formats`.

Be-ing:
MP3 patents have been expired for a few years now.

Be-ing:
https://en.wikipedia.org/wiki/MP3#Licensing,_ownership,_and_legislation

> The basic MP3 decoding and encoding technology is patent-free in the European Union, all patents having expired there by 2012 at the latest. In the United States, the technology became substantially patent-free on 16 April 2017

pdeljanov:
Yeah, I'm aware the patents have expired. However, it's still not a "free" codec in the true sense of the word (gotta pay for the specifications).

Beyond that, I believe it was a mistake to enable any codecs and formats by default at all. They all should've been all opt-in only. Then, a user who only wants one or two don't need to set `default-features = false`.

Likely I'll do that before 1.0.

Be-ing:
> They all should've been all opt-in only. Then, a user who only wants one or two don't need to set default-features = false.

Seems odd to me, but okay. :shrug:

PR #80, opened 2021-12-05
add mkv feature to symphonia crate

Be-ing:
This is listed in the README.md but the feature did not actually
exist.


pdeljanov:
The README notes that this is *In Work*. It's not ready to be enabled, or even merged for that matter. You can track progress in #66.

Be-ing:
That makes sense. In that case, I suggest to remove the documentation of the Cargo feature flag from README.md until the feature actually exists.

PR #81, opened 2021-12-05
add opus feature to symphonia crate

Be-ing:
This is listed in README.md but the feature did not actually exist.


pdeljanov:
The README notes that this is *In Work*. Opus support is basically nonexistent at this point and is a long-term goal of the project. Thus it's not ready to be added as a feature.

Issue #17, opened 2020-12-11
Basic examples


ndarilek:
Hello,

Found your library while surveying the landscape of Rust audio decoders. I'm trying to decode arbitrary audio codec data into raw bytes for use in Alto.

Unfortunately, I've spent about half an hour looking through the code and can't seem to figure out what my entrypoint should be. Mind pointing me to where I'd look to, say, ingest arbitrary bytes and decode them? Or, if I need to know the incoming content type first, where I'd go to do that? I'm working with a game engine that hands me raw bytes, though it also gives me a path from which I can perform reasonable content-type detection. I'm

Thanks!

ndarilek:
OK, looks like symphonia-play is what I need. Maybe a note in the README.md to that effect would help? The crate names aren't hugely helpful on their own. :)

I'll put together a PR for that soon if no one gets to it first.

pdeljanov:
Hey @ndarilek,

Glad you managed to figure it out. :smile: 

You're right, the README documentation is lacking. Definitely need to update the main README and probably all the others before the next version is published!

daniellga:
Hi! First of all, thank you for this amazing crate. I am trying to represent an audio example (flac) as a vec but I am having the same problem as the guy who created this issue. While in claxon I managed to achieve this reasonably fast, I am struggling to discover what to do here. Do you have any hints of what I can do to achieve this?

pdeljanov:
Hi @daniellga,

At the moment the current best example is `symphonia-play`. Unfortunately it has become a bit complex since it has to handle all the use-cases Symphonia supports.

Here's a [gist](https://gist.github.com/pdeljanov/3cc26a350c5fb0462c2fa65ffff27079) of the most basic decode loop one could implement. Note that there is no error handling!

I'll be adding this as an example in a bit. Based on the feedback in this ticket it sounds like more examples should be provided, but also probably a simpler interface that exposes a sample iterator.

PR #82, opened 2021-12-05
mention creek in README.md

Be-ing:


Be-ing:
I opened a [PR for creek](https://github.com/RustyDAW/creek/pull/6) to expose all of Symphonia's Cargo feature flags.

pdeljanov:
Thanks for updating Creek to support newer versions of Symphonia. I'm sure the project will benefit from the increased codec and format support.

That being said, this is an inappropriate pull request.

First, this is basically an advertisement placed in a prominent location in the README.

Second, Symphonia is perfectly safe and can be used as a standalone library as a developer sees fit.

Third, I am aware of, and have been contacted by the authors of, many libraries and projects that use Symphonia as a backend. Each of these projects have their own merits, and I do not wish to endorse any one project over the other.

Please don't make PRs like this.

Be-ing:
> Symphonia is perfectly safe and can be used as a standalone library as a developer sees fit.

Yes, it is, and that's great. Also, I think most developers considering using Symphonia would probably rather use a higher level library that allows them to simply specify a file path and get PCM audio to manipulate in their application or delegate to the library to simply play it for them. Looking at the [dependents on crates.io](https://crates.io/crates/symphonia/reverse_dependencies), there are only 4 crates depending on Symphonia currently:

  * creek: good for realtime audio applications
  * rodio: very high level API suitable for simple applications like a casual music player
  * kira: at a level of abstraction between creek and rodio and adds a few features for manipulating the audio, intended for use in games
  * babycat: a hodgepodge of features that brings in a C dependency... I don't quite get the purpose of this

pdeljanov:
I believe what you are asking for here are more examples. This is a fair critique. Ticket #17 is open if anyone wants to write and submit more examples to the project. Currently, `symphonia-play` is the main working example of using the library. Likewise, `symphonia-check` could also be used as an example as well.

> Looking at the [dependents on crates.io](https://crates.io/crates/symphonia/reverse_dependencies), there are only 4 crates depending on Symphonia currently:
> 
> * creek: good for realtime audio applications
> * rodio: very high level API suitable for simple applications like a casual music player
> * kira: at a level of abstraction between creek and rodio and adds a few features for manipulating the audio, intended for use in games
> * babycat: a hodgepodge of features that brings in a C dependency... I don't quite get the purpose of this

Yes, there are also other applications such as Ruffle, psst, and librespot (currently they're investigating usage) who are using Symphonia with their own audio pipelines that aren't on crates.io.

To be blunt, the main contention against this pull request is that you are trying to promote a project you contribute to in Symphonia's README specifically. Had you added a section towards the *end* of the README with a list of projects and libraries that use Symphonia I would've considered it. But again, this was basically promotion of a specific library.

Let's leave the issue at this.

Be-ing:
An example in the README would be helpful, sure. As someone looking at using this library, I would still appreciate links to other higher level libraries making use of Symphonia, and I think most developers considering the same would probably better off using those higher level libraries than reinventing code that already exists and works well.


Sorry for making a PR that frustrated you. Just for clarity, creek isn't my project; I've just made a few small pull requests to adapt it for [my use case](https://rust-audio.discourse.group/t/a-new-dj-application-in-rust/484) (which will likely result in building an even higher level library on top of creek for variable speed playback).

I hope we can work together on a better note going forward. Thank you very much for this really useful library. I am glad I don't need to bring in a dependency on FFmpeg in a Rust project.

pdeljanov:
Thanks for this. I'm willing to move past this issue as well. As you work on your project, feel free to raise any issues you have and let's see what we can do.

There's no fixed date for the next release, but I'm planning version 0.5 of Symphonia to be a "quality" release (although ALAC, and perhaps even MKV could be included). If you run into any snags or things that don't seem quite right, it can likely be fixed for that release.

PR #84, opened 2021-12-07
document that mkv and opus feature flags do not actually exist yet

Be-ing:


pdeljanov:
These tables are listed under the project roadmap with the status clearly indicated. The feature flags are listed there so other developers can know what to use if they decide to pick-up the feature.

The [docs.rs](https://docs.rs/symphonia/0.4.0/symphonia/) page does not list them.

You could remove the the lines from the table in `symphonia/README.md` instead since that's published on crates.io, but I want to keep them in the GitHub README for the reasons mentioned earlier.

Be-ing:
Okay, I reverted the edits to the root README.md and edited symphonia/README.md.

pdeljanov:
Merged, thanks!

PR #90, opened 2021-12-27
Derive Debug for Cue, Track, CuePoint, CodecParameters, VerificationCheck

antifuchs:
I'm writing software where having these data types be debug-printable would be helpful. I see some other types got `Debug` impls recently, so I hope this can make it in, too!

### DIFF
--- a/symphonia-core/src/codecs.rs
+++ b/symphonia-core/src/codecs.rs
@@ -229,7 +229,7 @@ pub const CODEC_TYPE_ALAC: CodecType             = CodecType(0x2003);
 pub const CODEC_TYPE_TTA: CodecType              = CodecType(0x2004);
 
 /// A method and expected value to perform verification on the decoded audio.
-#[derive(Copy, Clone)]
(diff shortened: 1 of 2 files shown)


pdeljanov:
Hey @antifuchs,

Thanks for this. Yeah, I see no problem with adding these.

According to the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/checklist.html) we should be eagerly deriving these traits for most types. Not deriving them seems to be a mistake.

antifuchs:
Brilliant, thank you! You're right that there are a few more that might be missing. I might be able to get the remainder decorated, too.

pdeljanov:
Sure, if you have time and want to go over the public types to make sure they're decorated, that'd be great! No worries if you can't though.

I noticed that `std` does implement `Debug` via. a custom implementation for  "class-like" structs such as `BufReader`, but it's kinda subjective regarding what should be printed so I don't think we should do that. I think if you just decorated public enums and the simple "data-only" structs we should be fine.

Thanks!

Issue #88, opened 2021-12-18
#![no_std] support


mooman219:
Are there plans to push for `#![no_std]` support in the future, specifically `core + alloc`? By the looks of it, there are a couple of std calls that for the most part have a 1:1 conversion, but path, file, and IO usually don't. Only including those via a flag would fill that gap.

pdeljanov:
It's a good idea and I'd support it if someone wants to take on the task, but I don't think it's something I'd take on for a long while.

There would also be some off-shoot work required to make Symphonia more viable for things like embedded targets. Things like optimizing bit IO for 32-bit machines, lowering memory usage, implementing more optimizations, etc.

mooman219:
I think I have time to put together some PRs to start the equivalent `std` -> `core` conversions. Changes after that would be marking off API that can't be converted as requiring `std`, and including that as a default feature flag as to minimize breaking changes to the API.

pdeljanov:
Hi @mooman219,

I probably should've read this before looking at your first PR.

I don't have any experience writing a no_std library so unfortunately I can't offer any solid advice or direction. I'd need to survey what other crates are doing first before I can form an opinion. Or, if you have some ideas or know what the best practice is, we can discuss them.

I do have a couple general thoughts based on the research I've done since reading your message.

1. I think we can work around the lack of `std::io` fairly easily. Symphonia relies on its own IO traits (e.g., `ReadBytes`), so filling in the gaps should be pretty simple albeit with some refactoring.

2. The larger challenge will be the lack of a `std::error::Error`. I don't currently have any ideas for this. Perhaps there's a crate that can cover it for us? However, for default/`std`-enabled builds, I think we want to continue using `std::error::Error` for consistency with the rest of the ecosystem. More generally, I don't want to add any complexity to default/`std`-enabled users since I do believe they will be the vast majority.

Going forward, perhaps we should merge any PRs into a new branch since these changes may take some time to land and settle. Alternatively, could also just push all the commits to the PR branch and merge when it's time.

mooman219:
Seems sensible, the io/error related changes are the biggest pain. There's the `core2` crate which explores the situation. There are also a couple of RFCs that try to address this issue, but they're quite old.

The history here is that _most_ of IO could be in core, but the error type couldn't be represented without an allocation, which at the time was more complicated to represent in core, so everything was left in std.

pdeljanov:
Hi @mooman219,

[...]

PR #91, opened 2021-12-28
Apply 1:1 conversions

mooman219:
Initial work for https://github.com/pdeljanov/Symphonia/issues/88

Apply equivalent 1 to 1 conversions. Dependencies `alloc` and `hashbrown` are added.

There is not default format config, `rustfmt.toml` on this project. Is that intentional? I'd be happy to re-format if you push some config you like.


pdeljanov:
Hi @mooman219,

Thanks for taking on this task. I'm sure there will be many people who appreciate this work!

Is there a way to do this conversion without having to invoke `rustfmt`?

The reason why there is no config is because I generally just don't want anyone to run `rustfmt` until I settle on a proper configuration. However, there are some challenges. First, many of the configuration options I'd like to use are nightly only (e.g., `control_brace_style`). Second, last time I checked there wasn't a very good way to vertically align things. Third, I need to go through the code and insert `#[rustfmt::skip]` where it belongs (such as on some big tables) which is a time consuming process.

Aside from that, having formatting changes in this PR also makes it difficult for me to review what has actually meaningfully changed.

Thanks!

mooman219:
Hey @pdeljanov,

That's reasonable, here is it without formatting. 

---
For fixed indentation I usually just format the code to force a desired wrapping:
```
pub const IDENTITY_MATRIX: Matrix4<f32> = Matrix4::new(
    1.0, 0.0, 0.0, 0.0, //
    0.0, 1.0, 0.0, 0.0, //
    0.0, 0.0, 1.0, 0.0, //
    0.0, 0.0, 0.0, 1.0, //
);
```

pdeljanov:
Hey @mooman219,

Thanks! This looks a fairly straightforward and clear change. Thanks for that tip as well, very clever!

I missed your last reply to the issue you raised for this. I've left a reply. For now I'll leave this PR open until we decide on the best way forward. :)

PR #93, opened 2022-01-04
core: derive `PartialEq` for standard key types

Techno-coder:
Hi, this small PR derives `PartialEq` for the standard key metadata types. I'd like this for searching for a particular tag dynamically.


pdeljanov:
Thanks @Techno-coder!

PR #98, opened 2022-01-22
Make FormatReader sync

sagudev:
similar as https://github.com/pdeljanov/Symphonia/issues/24 just for impl sync trait.

### DIFF
--- a/symphonia-codec-vorbis/src/floor.rs
+++ b/symphonia-codec-vorbis/src/floor.rs
@@ -111,7 +111,7 @@ macro_rules! try_or_ret {
     };
 }
 
-pub trait Floor: Send {
+pub trait Floor: Send + Sync {
     fn read_channel(
         &mut self,
         bs: &mut BitReaderRtl<'_>,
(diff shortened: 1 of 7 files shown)


sagudev:
Had some problems with VSCode formatter which used stable formatter.

The only active nightly formating is `control_brace_style = "ClosingNextLine"`. Is it really needed?

sagudev:
ping @pdeljanov

pdeljanov:
Hi @sagudev,

I'm not 100% up-to-speed on Send/Sync, so I will need to read up and consider the implications of this change and if it'll impose any restrictions to developers. My initial thought is that it should be fine.

That being said, this would be a semver breaking change so I'll make sure to do that before version 0.5 which should be in a couple weeks.

pdeljanov:
Hi  @sagudev,

Thank you for your patience. I think this is a good idea and I see little obvious downside. Would you mind also adding `Sync` to the `codecs::Decoder` and `meta::MetadataReader` traits for consistency and so we can get all the changes in a single commit?

Thanks!

sagudev:
I added Sync to all traits that already implemented Send.

pdeljanov:
Thank you @sagudev!

PR #103, opened 2022-01-29
Meta: Fix reading syncsafe integers with a bit width of 32

### LINKED ISSUE #102: `attempt to subtract with overflow` error when parsing id3 crc32
Running [this file](https://mega.nz/file/szRABbDb#1HvTGjQ-IrNGTlCPuSNkQ5BgFctejtW2LEWg02iUtWQ) with symphonia-play results in the following error: `thread 'main' panicked at 'attempt to subtract with overflow', symphonia-metadata/src/id3v2/unsync.rs:21:58`. The error comes from [this line](https://github.com/pdeljanov/Symphonia/blob/87eef4916ef976ea335e3051ac62a1187d1782f7/symphonia-metadata/src/id3v2/unsync.rs#L21) and particularly this bit: `(bit_width - bits_read)`. Most places call this function with a bit width of 28, which works okay because `bits_read` is incremented by 7 each time which divides evenly into 28. However, when this is called with a bit width of 32 on [this line](https://github.com/pdeljanov/Symphonia/blob/87eef4916ef976ea335e3051ac62a1187d1782f7/symphonia-metadata/src/id3v2/mod.rs#L215), it causes the while loop [here](https://github.com/pdeljanov/Symphonia/blob/87eef4916ef976ea335e3051ac62a1187d1782f7/symphonia-metadata/src/id3v2/unsync.rs#L19) to perform the following calculations for `(bit_width - bits_read)`:
```rust
32 - 7
32 - 14
32 - 21
32 - 28
32 - 35
```
Once it gets to the last line it overflows since it goes negative. I was able to fix this by changing the while loop to the following:
```rust
while bits_read < bit_width {
    // Ensure bits_read never exceeds the bit width which will cause an overflow
    bits_read = (bits_read + 7).min(bit_width);
    result |= u32::from(reader.read_u8()? & 0x7f) << (bit_width - bits_read);
}
```
After making this change, symphonia-play prints out all the tags as expected. If this is the correct approach, I'd be happy to make a PR for this. Just wanted to check if this is the correct way to fix it first. Thanks!

aschey:
Resolves #102 as discussed. Let me know if there are any other test cases I should account for, thanks!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -15,6 +15,7 @@ Thom Chiovoloni <chiovolonit@gmail.com>
 #
 # Please keep this section sorted in ascending order.
 
+aschey [https://github.com/aschey]
 BlackHoleFox [https://github.com/blackholefox]
 darksv [https://github.com/darksv]
 djugei [https://github.com/djugei]
--- a/symphonia-metadata/src/id3v2/unsync.rs
+++ b/symphonia-metadata/src/id3v2/unsync.rs
@@ -20,12 +20,12 @@ pub fn read_syncsafe_leq32<B: ReadBytes>(reader: &mut B, bit_width: u8) -> Resul
         // Ensure bits_read never exceeds the bit width which will cause an overflow
         let next_read = (bit_width - bits_read).min(7);
         bits_read += next_read;
-        // The second-last hexadecimal digit of the mask should be equal to the number of bits read
-        let mask = 0x0f + (next_read << 4);
+        // The mask should have of the bits below 2 ^ nex_read set to 1
+        let mask = (1 << next_read) - 1;
         result |= u32::from(reader.read_u8()? & mask) << (bit_width - bits_read);
     }
 
-    Ok(result & (0xffff_ffff >> (32 - bit_width)))
+    Ok(result)
 }
 
 pub fn decode_unsynchronisation(buf: &mut [u8]) -> &mut [u8] {
@@ -191,12 +191,15 @@ mod tests {
 
         // Special case: for a bit depth that is not a multiple of 7 such as 32
         // we need to ensure the mask is correct.
-        // In this case, the final iteration should read 4 bits and have a mask of 0x4f.
-        // 0x4f == 1001111 and has a 0 in 16's place so testing mask & 16 will ensure this is working.
+        // In this case, the final iteration should read 4 bits and have a mask of 0b0000_1111.
+        // 0b0000_1111 has a 0 in 16's place so testing mask & 16 will ensure this is working.
         let mut stream = BufReader::new(&[16, 16, 16, 16, 16]);
         assert_eq!(541098240, read_syncsafe_leq32(&mut stream, 32).unwrap());
 
         let mut stream = BufReader::new(&[3, 4, 80, 1]);
         assert_eq!(6367233, read_syncsafe_leq32(&mut stream, 28).unwrap());
+
+        let mut stream = BufReader::new(&[3, 4, 80, 1]);
+        assert_eq!(0, read_syncsafe_leq32(&mut stream, 0).unwrap());
     }
 }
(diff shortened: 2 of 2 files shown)


pdeljanov:
Thanks for the quick turnaround on this!

Unfortunately, I just noticed a second, unrelated to your change, issue with this function that's quite serious.

Each iteration of the loop reads a byte, masks off the upper-most bit, and then ORs it onto the result going from the most-significant bits down to the least. If `bit_width` is a multiple of 7, there is no problem since every read is using the whole byte (the mask is `0x7f`). However, if the last read uses less-than 7 bits, then the extra upper-most bits (which are controlled by the file) end up being OR'd into the result as well because the mask is still `0x7f`.

It looks like the final AND at the end of the function tried to account for that, but it seems we need to adjust the mask correctly for each read byte instead.

This scenario also brings up a good additional test-case that should be added. I'd probably also add a test for when `bit_width` is 0.

On another note, it looks like you've contributed to this project before but aren't in the `CONTRIBUTORS` file. Please consider adding yourself to it if you would like.

aschey:
@pdeljanov I just pushed a change that I think addresses your first point. Now if the number of bits read is 4 for example, the mask will be `0x4f` instead of `0x7f`. I added another test case to catch that since it only makes a difference on a subset of inputs. If this isn't the fix you had in mind, let me know and I can correct it. As far as accounting for a bit depth of 0, currently that will panic because `0xffff_ffff >> 32` causes an overflow. I'm not sure how that case should be handled, but let me know and I can add something to handle that edge case.

pdeljanov:
Hey, I think the general idea is right, but the mask generation looks off to me.

The goal is to only extra the lower `next_read`-bits from the read byte, which means the lower `next_read`-bits in the mask needs to be all 1s.

However, if we look at what the current code generates, we get something like this:

| `next_read` | `mask` (computed) | `mask` (required) |
|-------------|---------------------|-------------------|
| 0 | `0b0000_1111` | `0b0000_0000` |
| 0 | `0b0000_1111` | `0b0000_0000` |
| 0 | `0b0000_1111` | `0b0000_0000` |
| 0 | `0b0000_1111` | `0b0000_0000` |

pdeljanov:
Hey @aschey,

The general idea looks right, but the mask generation seems a bit off.

The mask needs to have the lower `next_read`-bits set to 1, however the current change doesn't compute that. Here's a table for comparison.

| `next_read` | `mask` (computed) | `mask` (required) |
|-------------|---------------------|-------------------|
| 0 | `0b0000_1111` | `0b0000_0000` |
| 1 | `0b0001_1111` | `0b0000_0001` |
| 2 | `0b0010_1111` | `0b0000_0011` |
| 3 | `0b0011_1111` | `0b0000_0111` |
| 4 | `0b0100_1111` | `0b0000_1111` |
| 5 | `0b0101_1111` | `0b0001_1111` |
| 6 | `0b0110_1111` | `0b0011_1111` |
| 7 | `0b01110_1111` | `0b0111_1111` |

The way to generate a mask where the lower `num` bits are `1` can be done by `let mask = (1 << num) - 1;`

As for the `0xffff_ffff >> 32`, that should not be required anymore because each byte is being masked correctly. In other words, `result` can just be returned without any further masking.

aschey:
@pdeljanov Thank you for adding the table, that definitely helped explain the concept. I believe everything should be working as intended now. I forgot to mention previously that I changed `bit_width` from a `u32` to a `u8` in order to prevent an extra cast in the mask calculation. Since the max value for `bit_width` that I see currently is 32, I don't think this should be an issue, but let me know if I'm missing something there and I can change it back.

pdeljanov (approved):

pdeljanov:
I don't think there's any problem with changing `bit_width` to `u8` so this all looks good to me. Thanks for the fixes!

Issue #43, opened 2021-07-11
symphonia-play crashes on Windows 10 if the audio file's sample rate doesn't match the output device's default sample rate 


Herohtar:
If I try to play an audio file that has a sample rate that is different than the sample rate that is set as the default format for my audio device, `symphonia-play` crashes. I'm guessing that this might not strictly be a bug, per se, but just a side effect of opening the default audio device with its default configuration. If so, I think to resolve this it would have to check the sample rate of the file before opening the device, then try to open the device using a matching sample rate. Is that correct?

```text
 ERROR symphonia_play::output::cpal > audio output stream open error: The requested stream configuration is not supported by the device.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OpenStreamError', symphonia-play\src\main.rs:248:74
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\release\symphonia-play.exe D:\test.flac` (exit code: 101)
```

pdeljanov:
Hmm, I don't think that's the issue because the audio device is opened [here](https://github.com/pdeljanov/Symphonia/blob/35accfec80d996abf05aeaa9e6a4f17e26b83535/symphonia-play/src/output.rs#L205-L209) which is only called by `symphonia-play` after the first audio buffer is decoded. So it should have the correct sample rate. This code is also used on macOS and it seems to work properly. :thinking: 

I think @BlackHoleFox had a similar issue. It may be device specific. Unfortunately I don't have a Windows machine easily accessible at the moment to debug this.

Perhaps try to check the sample rate (`spec.rate`) it's trying to open the device with?

Herohtar:
Ah, I was looking at the wrong thing. I checked, and `spec.rate` is the same rate that is listed in the printed track info, which is the sample rate that the file was exported as, so it's trying to open the device using the correct sample rate, but `build_output_stream` still returns that error.

It is definitely the sample rate that is causing the issue, because if I re-export the track with the same sample rate that I have set as my audio device's default output format, it will play. Alternatively, I can set my audio device's default output format to match the sample rate of the original track, and it will also play.

I suppose it could be that something about the audio device is not allowing it to be opened using a sample rate that isn't the default, though I'm not sure why. If it is something code-related, I guess it would be on `cpal`'s end though.

pdeljanov:
[...]

PR #109, opened 2022-02-17
core/audio/RawSampleBuffer: allow using external storage

fengalin:
@pdeljanov: many thanks for the awesome crates!

I'm working on a Symphonia based GStreamer plugin ([WIP branch](https://gitlab.freedesktop.org/fengalin/gst-plugins-rs/-/tree/symphonia)). So far, I've integrated the FLAC and MP3 decoders, which can already be used in pipelines as drop-in replacements for any other similar decoders. My plan would be to integrate the other coders with "excellent" status and probably demuxers too. But before going any further, I'd like to get your feeling about this.

I'd also want to propose a modification that would make the integration a bit more optimal IMHO. Quoting the commit message:

> GStreamer uses its own Buffers which are handed to the downstream
nodes in the processing graph. The nodes (elements) can be implemented
in mutiple languages and the Buffers can be allocated elsewhere or be
part of a Buffer pool.
> 
> RawSampleBuffer provides a convenient mechanism to export decoded
buffers to an unaligned bytes stream. In order to ensure data
safety, the RawSampleBuffer manages its own sample storage and
provides an accessor to obtain an immutable byte slice on the
resulting data.
> 
> When used in frameworks such as GStreamer, the byte slice must be
copied to the target Buffer.
> 
> The changes in this commit allow using the RawSampleBuffer with
an external byte slice, thus avoiding the last copy.

I left open questions in the code (see `FIXME`s). After a back and forth, I decided to use an `&mut [u8]` for the external storage type, but I also considered `&mut [S::RawType]`. Reasons why I settle to `&mut [u8]` are:

- We can still ensure safety by carefully checking the available size.
- It's more convenient for callers: otherwise, they would most likely need to cast their slice using an external crate and `Symphonia` already makes use of such a crate.
- This is symmetric to the output type returned by `RawSampleBuffer::as_bytes`.

Is this acceptable to you?

### DIFF
--- a/symphonia-core/src/audio.rs
+++ b/symphonia-core/src/audio.rs
@@ -956,20 +956,64 @@ impl RawSample for f64 {
     }
 }
 
+/// Raw Samples owned or borrowed Storage.
+enum RawSampleStorage<'a, S: Sample + RawSample> {
+    Owned(Vec<S::RawType>),
+    Borrowed(&'a mut [u8])
+}
+
+impl<'a, S: Sample + RawSample> RawSampleStorage<'a, S> {
+    /// Returns an immutable slice containing the entire storage content.
+    fn as_slice(&self) -> &[S::RawType] {
+        use RawSampleStorage::*;
+        match self {
+            Owned(owned) => owned.as_slice(),
+            Borrowed(slice) => bytemuck::cast_slice(slice),
+        }
+    }
+
+    /// Returns a mutable slice containing the entire storage content.
+    fn as_mut_slice(&mut self) -> &mut [S::RawType] {
+        use RawSampleStorage::*;
+        match self {
+            Owned(ref mut owned) => owned.as_mut_slice(),
+            Borrowed(ref mut slice) => bytemuck::cast_slice_mut(slice),
+        }
+    }
+}
+
+impl<S: Sample + RawSample> From<Vec<S::RawType>> for RawSampleStorage<'static, S> {
+    fn from(owned: Vec<S::RawType>) -> Self {
+        RawSampleStorage::Owned(owned)
+    }
+}
+
+impl<'a, S: Sample + RawSample> From<&'a mut [u8]> for RawSampleStorage<'a, S> {
+    fn from(borrowed: &'a mut [u8]) -> Self {
+        assert_eq!(
+            borrowed.len() % mem::size_of::<S::RawType>(),
+            0,
+            "inconsistent borrowed storage size",
+        );
+
+        RawSampleStorage::Borrowed(borrowed)
+    }
+}
+
 /// A `RawSampleBuffer`, is a byte-oriented sample buffer. All samples copied to this buffer are
 /// converted into their packed data-type and stored as a stream of bytes. `RawSampleBuffer` is
 /// mean't for safely importing and exporting sample data to and from Symphonia as raw bytes.
-pub struct RawSampleBuffer<S: Sample + RawSample> {
-    buf: Vec<S::RawType>,
+pub struct RawSampleBuffer<'a, S: Sample + RawSample> {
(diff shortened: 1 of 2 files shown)


pdeljanov:
Hi @fengalin,

This would be a nice improvement to `RawSampleBuffer` since it does save a copy. I also see this being useful for `SampleBuffer` since it suffers from the same problem.

I will have to take a closer look at this once I get more time, but I do see one issue. 

`RawSampleStorage::as_mut_slice` calls `bytemuck::cast_slice_mut` to convert the `&mut [u8]` to a `&mut [S::RawType]`. This function may panic if the alignment of `S::RawType` is greater than the alignment of `u8`. Since the alignment of a `u8` is `1`, this is usually always true. The reason why this did not panic for you is because the Rust allocator on mainstream operating systems aligns all allocations to 16 byte boundaries. So there is an exception to the aforementioned rule if the pointer underlying the `&mut [u8]` is a multiple of the alignment of `S::RawType`. Since every `S::RawType` has an alignment less-than 16 bytes, and all allocations are aligned to 16 byte boundaries, the panic will never occur. Therefore, in practice there is no issue on these systems. However, this is not guaranteed and this could cause panics on other systems or if the allocator is changed. You can see how bytemuck checks for this [here](https://docs.rs/bytemuck/latest/src/bytemuck/lib.rs.html#448-471).

Fundamentally, this is the reason why `RawSampleBuffer` allocated a `[S::RawType]` and converted it to a `[u8]`. It can never fail, because a `u8` has the least strict alignment of all types.

I guess we could use `TryFrom` instead of `From` for `RawSampleStorage` and do a check there?

pdeljanov:
On a different note. Since this change will cause a semver breaking change, and version 0.5 was just released, it may be a while before we can release a 0.6 since I want to roll-up a bunch of other not-yet-started breaking changes in 0.6.

fengalin:
Almost forgot: there's something else I need to address. In existing GStreamer FLAC and Vorbis decoders, there is a mapping between the codec's channel positions and the framework's conventions. I think we could also add a channel mapping feature to `[Raw]SampleBuffer`.

pdeljanov:
Hi @fengalin,

Thanks for your work on this. Please give me some time to go over the changes more thoroughly, unfortunately I'm a bit short on time at the moment!

> Almost forgot: there's something else I need to address. In existing GStreamer Vorbis decoders, there is a mapping between the codec's channel positions and the framework's conventions. I think we could also add a channel mapping feature to `[Raw]SampleBuffer`.

Does GStreamer output in the native Vorbis channel order? Symphonia outputs in the [`WAVEFORMATEXTENSIBLE`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ksmedia/ns-ksmedia-waveformatextensible) order which is pretty typical I believe.

That being said, channel mapping(/mixing) was brought up in #43, and was a feature I had in my todo list. My idea was that on any audio buffer copy an optional channel matrix could be provided and the framework would take care of remapping and/or mixing the channels during the copy. However, I'm now thinking that it would be better for performance and source compatibility if separate copy functions are provided for such cases.

fengalin:
> unfortunately I'm a bit short on time at the moment!

No worries! I had this integration in the pipeline since more than a year, so it can wait all right!

Just to clarify (and hopefully simplify) the channel mapping stuff: I was only referring to re-ordering identical number of channels. Channel mixing would not be part of the requirements for a GStreamer audio decoder as this is handled by dedicated elements. Anyway, I still need to dig a bit more in the Vorbis implementations to make sure I really need this, so let's just keep it aside for the moment. :)

pdeljanov:
A pre-release version 0.6 is due to be released very soon. In 0.6 the problem has been solved by providing methods on audio buffers & slices to copy the audio data in a myriad of different ways. Thanks for your input on this!

PR #111, opened 2022-02-21
opus: add initial support (WIP)

thinking-tower:
@pdeljanov, could you re-checkout `opus-dev` from master? I've nuked my branch because it duplicated what was already done in `symphonia-format-ogg/src/mappings/opus.rs`.

@darksv pinging you so you can get an idea of where the Opus progress is at. 

# Description
Part of the work for #8.

Stages:
- Implementing [RFC 6716 Section 3.1](https://datatracker.ietf.org/doc/pdf/rfc6716).


pdeljanov:
Hi @thinking-tower,

Are you sure about that? The code in the file you mentioned skips information that's not required for the `OggReader` such as the gain. I imagine `OpusDecoder` will want such information.

I could rebase `opus-dev` to the latest `master` so you don't lose your existing work. Actually, I was hoping that the extra data parser your wrote for Opus could be moved to `symphonia-utils-xiph` and used by `symphonia-format-ogg` in the future.

thinking-tower:
@pdeljanov 

> Are you sure about that? The code in the file you mentioned skips information that's not required for the OggReader such as the gain. I imagine OpusDecoder will want such information.

> Actually, I was hoping that the extra data parser your wrote for Opus could be moved to symphonia-utils-xiph and used by symphonia-format-ogg in the future.

Good point! Updating the parser can be deferred until I need it in the decoder. 

> I could rebase opus-dev to the latest master so you don't lose your existing work.

That'll be awesome!

pdeljanov:
Done! I force pushed the rebased branch to `opus-dev`. I'll update the base branch for this PR to `opus-dev`.

thinking-tower:
@pdeljanov I've left the header code in there for now, there will be opportunities to clean it up once it's clearer how it should be used for the decoder.

pdeljanov:
Merged, thanks!

PR #135, opened 2022-06-07
core: Prioritize formats that match the hint when determining the container format

### LINKED ISSUE #130: Format probe incorrectly detects mp3 file as aac
When attempting to play [this file](https://mega.nz/file/IqgmjTwS#DMfxlYECsnmAjX9t_roTGMDLWij5BrH_cr4r-6w8-po) with `symphonia-play`, the following error is returned: ` ERROR symphonia_play > file not supported. reason? unsupported feature: adts: only 1 aac frame per adts packet is supported`. This is because the probe detects an aac marker [when it gets to the code here](https://github.com/pdeljanov/Symphonia/blob/41862b26699d6474c4e5d8415c50f1eab0972fe1/symphonia-core/src/probe.rs#L269), but the file is actually an mp3. When I disable the aac feature flag, the file plays fine. 

Seems like the easiest way to get around this would be to [pass along the hint](https://github.com/pdeljanov/Symphonia/blob/41862b26699d6474c4e5d8415c50f1eab0972fe1/symphonia-core/src/probe.rs#L305) to the `next` function and ensure that it checks the codec that corresponds to the hint before checking the other codecs, which should prevent false positives in the common case where the filetype is known. I guess a more robust way would be to make the markers stricter if possible or attempt to decode the first bit of the file before assuming the marker is correct, but that sounds messy. Curious what your thoughts are, thanks!

aschey:
Resolves #130 

This PR changes the format detection behavior so that all formats that match the supplied extension or mime type are checked first. Some notes on the implementation:

- My initial idea was to keep the original structure as similar as possible and just change [the inner loop](https://github.com/pdeljanov/Symphonia/blob/050c89210550b0883ff24d5df5740f85d818d216/symphonia-core/src/probe.rs#L264) to prioritize the formats matching the hint, but the file I was testing with incorrectly matched the AAC marker on the second call to `next()` and only found the correct mp3 marker on the fifth call. Therefore, the only way to prevent false positives is to run through the entire file using the formats that match the hint before falling back to the other formats.
- This will have a slight performance hit in the case where an incorrect hint is provided, but should have a negligible impact if no hint is supplied.
- The existing log messages seem to suggest that `Probe::PROBE_SEARCH_LIMIT` should be the max number of bytes that are checked when determining the format, but I didn't see this constant being checked in the code anywhere. This change enforces that no more than `Probe::PROBE_SEARCH_LIMIT` bytes are checked. Enforcing this was necessary because the case where an incorrect hint is supplied became very slow without enforcing this limit.
- Rather than call `to_lowercase()` every time we need to check the extension or mime type, I opted to normalize it when the value is initially set. Considering all the extensions and mime times referenced in the code are all lowercase, I didn't see a need to store any non-lowercase version that the user supplied.

Let me know if you'd like me to change anything, thanks!

### DIFF
--- a/symphonia-core/src/probe.rs
+++ b/symphonia-core/src/probe.rs
@@ -14,6 +14,7 @@ use crate::io::{MediaSourceStream, ReadBytes, SeekBuffered};
 use crate::meta::{Metadata, MetadataLog, MetadataOptions, MetadataReader};
 
 use log::{error, info};
+use std::io::{Seek, SeekFrom};
 
 mod bloom {
 
@@ -125,6 +126,24 @@ pub trait QueryDescriptor {
     fn score(context: &[u8]) -> u8;
 }
 
+/// `ContainsOptional<T>` is a trait to define the behavior of checking whether a collection-like
+/// type contains an element which may be `None`.
+trait ContainsOptional<T>
+where
+    Self: IntoIterator,
+{
+    fn contains_optional(&self, search: &Option<T>) -> bool;
+}
+
+impl ContainsOptional<String> for &[&str] {
+    fn contains_optional(&self, search: &Option<String>) -> bool {
+        match search {
+            Some(str) => self.contains(&&str[..]),
+            None => false,
+        }
+    }
+}
+
 /// A `Hint` provides additional information and context when probing a media source stream.
 ///
 /// For example, the `Probe` cannot examine the extension or mime-type of the media because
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hi @aschey,

Thanks for prototyping this. I think you hit a case that I may not have considered during our earlier discussion.

Unfortunately, I don't think the approach of scanning through the file (or a portion of it) first for the hinted format is going to be acceptable.

There are a few problems with doing that:
1. If the `MediaSourceStream` is not seekable then you are limited to the size of the media source stream's internal buffer. This buffer is resizeable, but it shouldn't grow too large (generally no more than 64kB).
2. However, even if you do grow the buffer to the maximum search size, it'll then impose a large  latency for streaming sources (all those bytes would need to be buffered first which could be 10s of seconds of audio).
3. For things like MP3 files, they usually have and ID3v2 tag tacked onto them at the start of the file which has a different marker than what the MP3 reader looks for. This is why `Probe` is iterator-like. It first yields a `Id3v2Reader` and then a `Mp3Reader`. Unfortunately this means when a MP3 is hinted, it'll almost always pay the cost of a full search.

Because of this issues, I think whatever algorithm is used it has to work progressively through the stream.

FelixMcFelix:
I looked into trying this a while ago when I was digging into writing a container format for local use -- I ended up running into the same ID3 issue with MP3s.

Is a possible solution that some containers should have 'dependencies' on other formats or metadata? I.e., an MP3 MIME tag suggests the format, but the dependency tells the `Probe` to look for an ID3v2 tag first. At the same time, in the case of something like an MKV you might be able to ignore probe metadata like ID3 tags (hopefully saving some time in the probe step).

pdeljanov:
Hi @FelixMcFelix,

That's an idea, but I've also seen some weird combinations that would normally be unexpected. For example, I have some FLAC files which have ID3v2 tags. In the end we would likely have to look for ID3v2 tags on everything! There's also the same chance of hitting a false positive match.

aschey:
I'll go ahead and close this since it sounds like this approach won't work. Thanks for taking a look.

PR #137, opened 2022-06-15
ogg: Apply TimeBase to contained Opus streams

FelixMcFelix:
This allows users to convert seek results for Ogg Opus files given in `TimeStamp`s back into `Time`s without having to track special cases.


pdeljanov:
Nice and simple, I like it! :)

PR #138, opened 2022-06-15
mkv: Fix forward seeks with read-only MediaSources

FelixMcFelix:
Read-only sources were mistakenly calling `Seek::seek` if an MkvReader's `clusters` had been populated, causing a seek error even if the target timestamp occurs later in the stream.

This PR changes the EBML iterator to consume bytes if the source is not seekable *and* the destination can be forward-skipped to, otherwise returning `ForwardOnly`.

### DIFF
--- a/symphonia-format-mkv/src/demuxer.rs
+++ b/symphonia-format-mkv/src/demuxer.rs
@@ -184,7 +184,7 @@ impl MkvReader {
                 Some(block) => block.pos,
                 None => cluster.pos,
             };
-            self.iter.seek(pos)?;
+            self.iter.seek(pos, self.iter.inner().is_seekable())?;
 
             // Restore cluster's metadata
             self.current_cluster =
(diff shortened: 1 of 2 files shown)


pdeljanov:
Hi @FelixMcFelix,

Sorry for the late response to this. The general idea looks good, but wouldn't it make sense to have the `ElementIterator` just check if the reader is seekable when seek is called rather than having the caller access the inner reader on the `ElementIterator` itself to get it's seekable status?

Perhaps the iterator could cache the value of `is_seekable` when the iterator is created?

FelixMcFelix:
I think at the time I was considering doing so, but wasn't sure of the best way to do it on account of the `R: ReadBytes` bound. We could either pass the value in via `ElementIterator::new(...)` or change the generic bound on `ElementIterator` to `R: ReadBytes + MediaSource`.

I think I was reluctant to change the generic bound at the time in case there was some extra intent I was missing, but I think it'd be clearer to change the [bound on seek](https://github.com/pdeljanov/Symphonia/blob/5daf53ed7aa1c195867fa41011222e11f3aa5f30/symphonia-format-mkv/src/ebml.rs#L254-L257) from `R: Seek` to `R: MediaSource`, which is technically breaking. What do you think?

pdeljanov:
Thank you @FelixMcFelix, that was a very elegant solution.

PR #141, opened 2022-07-03
isomp4: Fix incorrect parsing of SortArtist

101100:
The standard tag used was `Artist` when it should be `SortArtist`.


pdeljanov:
Thanks, nice catch!

PR #143, opened 2022-07-18
Replace lazy_static with phf and once_cell

GnomedDev:
- Reduces macro magic in the case of replacing with `once_cell::sync::Lazy` and removes a lot of lines due to no indentation required.
- Improves runtime performance in the case of replacing lazy_static'd `HashMap` with `phf::Map` by creating the maps at compile time, avoiding even one allocation at runtime


pdeljanov:
Hi @GnomedDev,

It looks like you put quite a lot of work into this. Unfortunately, I see some problems that give me pause:

1. Regarding the replacement of `lazy_static` with `once_cell`. According to the `once_cell` docs, `lazy_static` has an advantage in terms of supporting `#![no_std]`. I think we do want to eventually support no_std and we having a tracking ticket here #88. Are there any other benefits besides code brevity that outweigh this drawback?
2. In terms `phf` there are two issues I see:
     -  First, `phf` requires a MSRV of 1.60, while we support down to 1.53. Rust 1.60 is rather new. I was thinking of moving to 1.56 next. Either way, increasing the MSRV would be done for v0.6.
     - Second, since the hash tables are being generated at compile time, how does that affect binary size vs. the existing solution? We've already had a few issues raised regarding binary size and I'd prefer trading a few CPU cycles for a smaller binary.

GnomedDev:
Hey, sorry for completely missing your response.
1. once_cell is now a part of the Rust standard library, so the library should probably move to that if a breaking change is in order
2. I personally think if this change would be made, it move to a simple match statement instead of dragging in a dependency, if possible.

Since this has been left stale and is now conflicting so much, I will be closing this PR, feel free to use the ideas or code for a future PR.

edwloef:
The `once_cell` docs no longer recommend `lazy_static`, and claim to support `no_std` via the `race` module. Additionally, most projects seem to have made the transition to `once_cell` by now. Should this be given another go?

PR #144, opened 2022-07-18
Fix clippy warnings

GnomedDev:
This fixes existing clippy warnings, with a lot of them triggered by test code.

I guess clippy wasn't run before to find these test warnings, and then the existing warnings were hidden for users like me using rust-analyzer

### DIFF
--- a/symphonia-bundle-mp3/src/synthesis.rs
+++ b/symphonia-bundle-mp3/src/synthesis.rs
@@ -852,16 +852,17 @@ mod tests {
     use std::f64;
 
     fn dct32_analytical(x: &[f32; 32]) -> [f32; 32] {
-        const PI_32: f64 = f64::consts::PI / 32.0;
+        const PI_32: f32 = (f64::consts::PI / 32.0) as f32;
 
         let mut result = [0f32; 32];
-        for i in 0..32 {
-            let mut sum = 0.0;
-            for j in 0..32 {
-                sum += (x[j] as f64) * (PI_32 * (i as f64) * ((j as f64) + 0.5)).cos();
-            }
-            result[i] = sum as f32;
+        for (i, item) in result.iter_mut().enumerate() {
+            *item = x
+                .iter()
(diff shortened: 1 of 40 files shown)


symphonia-bundle-mp3/src/synthesis.rs
```diff
@@ -850,16 +850,17 @@ mod tests {
     use std::f64;
 
     fn dct32_analytical(x: &[f32; 32]) -> [f32; 32] {
-        const PI_32: f64 = f64::consts::PI / 32.0;
+        const PI_32: f32 = (f64::consts::PI / 32.0) as f32;
```
pdeljanov, line 853: Please don't change `f64` to `f32` in this test. We want as much accuracy as possible in the iterations that follow before returning an `f32`.

pdeljanov (changes requested):
Would be happy to accept these fixes. Just please don't reduce the accuracy of the analytical calculations for tests.

symphonia-core/src/conv.rs
```diff
@@ -615,7 +615,7 @@ impl_convert!(f32, f64, s, s as f64); // f64
 impl_convert!(f64, u8, s, ((s.clamped() + 1.0) * 128.0) as u8); // u8
 impl_convert!(f64, u16, s, ((s.clamped() + 1.0) * 32_768.0) as u16); // u16
 impl_convert!(f64, u24, s, u24::from(((s.clamped() + 1.0) * 8_388_608.0) as u32)); // u24
-impl_convert!(f64, u32, s, ((s.clamped() + 1.0) as f64 * 2_147_483_648.0) as u32); // u32
+impl_convert!(f64, u32, s, (s.clamped() + 1.0 * 2_147_483_648.0) as u32); // u32
```
pdeljanov, line 618: Removing the brackets changes the order-of-operations and results in a test failure.

pdeljanov:
Thanks for fixing all of these!

PR #145, opened 2022-07-18
Use NonZeroU32 in TimeBase

GnomedDev:
This removes unnecessary `asserts` in the methods, and allows for niche optimization.

I had to remove the `Default` implementation from TimeBase, but that makes no sense anyway as it just created a `TimeBase` which panics whenever used.

### DIFF
--- a/symphonia-core/src/units.rs
+++ b/symphonia-core/src/units.rs
@@ -7,7 +7,7 @@
 
 //! The `units` module provides definitions for common units.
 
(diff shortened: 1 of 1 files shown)


pdeljanov:
Sorry for the delay. Thanks for this, it makes sense to me.

Since the struct fields are public however, we will need to wait until v0.6 to merge this since it's a breaking change to switch from `u32` to `NonZeroU32`.

PR #147, opened 2022-07-25
core: Add `SampleBuffer::samples_mut`

FelixMcFelix:
Allows mutable access to exported samples, enabling in-place processing of output audio data.

Currently, to use functions like libopus's soft_clip (which take `&mut [f32]`, interleaved) this can only be achieved on `SampleBuffer` with a significant memcopy (seemingly ~8us for 1920 `f32` samples, relative to ~40us decode + ~80us encode). `samples_mut` removes this additional overhead.

### DIFF
--- a/symphonia-core/src/audio.rs
+++ b/symphonia-core/src/audio.rs
@@ -733,6 +733,11 @@ impl<S: Sample> SampleBuffer<S> {
         &self.buf[..self.n_written]
     }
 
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks @FelixMcFelix, this change makes a lot of sense to me. Apologies for the delay in getting this merged, got a bit too busy!

Issue #16, opened 2020-09-22
read_id3v2 cannot read chapter markers from file


rtyler:
I could have missed where this data might show up, but taking some sample podcasts from [Mac Power Users](https://www.relay.fm/mpu). (e.g. https://www.podtrac.com/pts/redirect.mp3/traffic.libsyn.com/secure/relaympu/mpu554.mp3)

Dumping out tags from the `Metadata` gives me this
```
k: TALB, v: Mac Power Users
k: TPE1, v: Mac Power Users
k: TIT2, v: 554: Read-it-later Services
k: COMM!eng, v: Read-it-later services can be a great way to save and enjoy an article later, away from the noise of social media or an overflowing RSS client. This week, David and Stephen talk about some of the popular choices, and how to keep them from becoming just another thing to check. T
hen, a recap of Apple’s recent media event.
k: USLT!eng, v: Read-it-later services can be a great way to save and enjoy an article later, away from the noise of social media or an overflowing RSS client. This week, David and Stephen talk about some of the popular choices, and how to keep them from becoming just another thing to check. T
hen, a recap of Apple’s recent media event.
k: TLEN, v: 5997000
k: TYER, v: 2020
k: TENC, v: Forecast
```

Whereas ffprobe on the same file will output: 

```
Input #0, mp3, from 'mpu554.mp3':
  Metadata:
    album           : Mac Power Users
    artist          : Mac Power Users
    title           : 554: Read-it-later Services
    comment         : Read-it-later services can be a great way to save and enjoy an article later, away from the noise of social media or an overflowing RSS client. This week, David and Stephen talk about some of the popular choices, and how to keep them from becoming just an
    lyrics-eng      : Read-it-later services can be a great way to save and enjoy an article later, away from the noise of social media or an overflowing RSS client. This week, David and Stephen talk about some of the popular choices, and how to keep them from becoming just an
    TLEN            : 5997000
    encoded_by      : Forecast
    date            : 2020
  Duration: 01:39:57.11, start: 0.000000, bitrate: 128 kb/s
    Chapter #0:0: start 0.000000, end 466.287000
    Metadata:
      title           : MPU 554
    Chapter #0:1: start 466.287000, end 964.340000
    Metadata:
      title           : Read-it-Later Services?
    Chapter #0:2: start 964.340000, end 1597.615000
    Metadata:
      title           : Safari Reading List
    Chapter #0:3: start 1597.615000, end 3042.409000
    Metadata:
[...]

PR #157, opened 2022-10-08
meta: Add support for id3v2 CHAP and CTOC frames

Tommoa:
These were standardised in https://id3.org/id3v2-chapters-1.0.

Something I initially considered (and did) was to use `Cue`s and `CuePoint`s instead of creating new types, but decided against it because:

- `Cue` is tied to the format, whereas these are tied to the metadata.
- According to the spec, chapters can overlap, and so must have a duration.
- It is unclear to me what the index should be.
- `CuePoint`s cannot have another `CuePoint` within them, which may be an issue for multi-level tables of content.
- Chapters and tables of contents would end up mixed as `Cue`s despite being different types.

Resolves: #16


pdeljanov:
Hi @Tommoa,

Thank you for implementing this!

I think you touch on a larger point that cues/chapters should be part of the metadata and not the format. I generally agree with this line of thinking. It's also basically why I avoided this issue for the time being.

I'd like to use your code, but I also think those larger API issues should be addressed first. After the v0.5.2 release I hope to get started on v0.6 where we can start making breaking API changes. Hopefully we can land on something good.

Thanks for your patience!

pdeljanov:
Hello, it's been a while!

I've rewritten Symphonia's chapter (formerly cue) support in the `dev-0.6` branch in dc287cc0f40c362d506132e2ee8689d7e9c91b16. In the process, I've implemented parts of this PR.

Since I referenced some of your code, I've added you to the `CONTRIBUTORS` file. Thanks again!

Issue #41, opened 2021-07-09
ADPCM support


thomcc:
I'd like to use Symphonia with some WAV files containing ADPCM data. Specifically, I'd like to support the "Microsoft ADPCM" (RIFF format 0x02<sup>1</sup>) and "IMA/DVI ADPCM" (RIFF format 0x11) variants. There are also possibly a couple extensions<sup>2</sup> that I need to support.

Just these two account for most of the ADPCM I've encounted in the wild, and to some extent rounds out the set of codecs that you most commonly find in WAV files, IME. The various other variants use can of course be added later when/if needed — I think there are a couple that likely will be useful (like QuickTime's "IMA4 ADPCM").

Anyway, I'd be willing to provide the code for those two (I have some parsing code lying around for them, and encoding is easy enough as well).

The reason I'm filing an issue first (rather than just a PR) is:
1. To make sure you'd accept it.
2. I am unsure where they should live. Or rather, I suspect the answer is "A new `symphonia-codec-adpcm` crate, but it seems worth asking about this first.

    I think there's a strong possibility you considered the existence of ADPCM when writing the WAV code, and possibly have thoughts on where it should live, but I could be wrong
3. So I have somewhere to ask questions when I inevitably get lost due to my vague-at-best understanding of how Symphonia is structured 😅

One final note is that while `libavcodec` (codec library behind ffmpeg)'s support for ADPCM is pretty thorough, `ffmpeg`'s usage of it used to introduce many strange bugs. So, I'm a little concerned about testing against it as with symphonia-check... I guess we'll cross that it it becomes an issue, though, since it may have been fixed in the 3 years since I tried last.

---
(footnotes)

1. I'm specifying the format ID because pretty much every name you could use to describe the different flavors of ADPCM is somewhat ambiguous. There are at least 3 codecs reasonably called "DVI ADPCM" and two reasonably called "Microsoft ADPCM", and these have overlap (it's kind of a disaster).

2. Specifically, the two important extensions are:
    -  Non-hardcoded block sizes, as documented https://docs.microsoft.com/en-us/windows/win32/xaudio2/adpcm-overview.
    - Arbitrary channel counts — I believe think the format inside WAV only supports 2 channels, according to some documentation. In practice I've seen upwards of 8, which decoded fine in windows and on macOS (using AVFoundation.

    Note that I'm unsure that these are actually extensions.

pdeljanov:
> 1. To make sure you'd accept it.

Yep.

[...]

Issue #124, opened 2022-05-09
Valid files(checked by FFprobe) doesn't properly parse in Symphonia


qarmin:
This files fails to open with latest git symphonia version - https://github.com/pdeljanov/Symphonia/commit/8f4aaed599ba8c23aab55d1bdad65ed621a68b92

FFprobe doesn't show any error with them

Code used to parse files - https://github.com/qarmin/audio_checker/blob/master/src/lib.rs

Example files
[5 - The Music Of Stasis - Stars And The Moons - Meesah Kuteyi feat Mark Morgan.mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655190/5.-.The.Music.Of.Stasis.-.Stars.And.The.Moons.-.Meesah.Kuteyi.feat.Mark.Morgan.mp3.zip)
[08 - Crush.mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655191/08.-.Crush.mp3.zip)
[idw_testsound.wav.zip](https://github.com/pdeljanov/Symphonia/files/8655193/idw_testsound.wav.zip)
[InFlux Original Soundtrack - 07 - Volcano Dance.mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655194/InFlux.Original.Soundtrack.-.07.-.Volcano.Dance.mp3.zip)
[jetpack1.wav.zip](https://github.com/pdeljanov/Symphonia/files/8655197/jetpack1.wav.zip)
[King of Dragon Pass - 24 - Danger.mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655199/King.of.Dragon.Pass.-.24.-.Danger.mp3.zip)
[Muzyka.ac3.zip](https://github.com/pdeljanov/Symphonia/files/8655200/Muzyka.ac3.zip)
[Neverwinter Nights 2 Storm of Zehir soundtrack - maintheme.mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655201/Neverwinter.Nights.2.Storm.of.Zehir.soundtrack.-.maintheme.mp3.zip)
[Robin Cook - Pandemia (54).mp3.zip](https://github.com/pdeljanov/Symphonia/files/8655202/Robin.Cook.-.Pandemia.54.mp3.zip)
[sample.wav.zip](https://github.com/pdeljanov/Symphonia/files/8655203/sample.wav.zip)
[...]

PR #160, opened 2022-11-04
Add ADPCM Decoder

geckoxx:
This adds an ADPCM decoder with following limitations:

- only Microsoft ADPCM and IMA ADPCM codecs
- only mono and stereo channels
- only 4bits per sample (IMA ADPCM allows 3bits)

It also expands the Wave Format to support ADPCM and the packetization over a number of samples (PCM) or blocks (ADPCM).
To support multiple data blocks per packet `CodecParameters` has a new parameter: `frames_per_block`.

Related Issue: #124 
Resolves: #41


symphonia-codec-adpcm/src/codec_ms.rs
```diff
@@ -0,0 +1,167 @@
...
+            let coeff_num = reader.read_u16()? as usize;
+            params.coeffs1.resize(coeff_num, 0);
+            params.coeffs2.resize(coeff_num, 0);
+            for i in 0..coeff_num {
+                params.coeffs1[i] = i16_to_i32!(reader.read_u16()?);
+                params.coeffs2[i] = i16_to_i32!(reader.read_u16()?);
+            }
+        }
+        Ok(params)
+    }
```
pdeljanov, line 58: Is this part needed? I'm getting conflicting information.

According [Microsoft](https://learn.microsoft.com/en-us/previous-versions/ee415686(v=vs.85)), the Microsoft ADPCM format only uses the built-in coefficients. However [this](https://wiki.multimedia.cx/index.php/Microsoft_ADPCM) reference mentions that additional coefficients can be *appended*. This implementation replaces the coefficients.

If MS ADPCM doesn't actually use this, then it's safer to remove this code since it's unclear what the correct  behaviour is.

symphonia-codec-adpcm/src/codec_ima.rs
```diff
@@ -0,0 +1,92 @@
...
+];
+
+/// `AdpcmImaBlockStatus` contains values to decode a block
+struct AdpcmImaBlockStatus {
+    predictor: i32,
+    step_index: i32,
+}
+
+impl AdpcmImaBlockStatus {
+    fn read_preample<B: ReadBytes>(stream: &mut B) -> Result<Self> {
```
pdeljanov, line 33: Should be this called `read_preamble` instead?

symphonia-codec-adpcm/src/codec_ms.rs
```diff
@@ -0,0 +1,167 @@
...
+struct AdpcmMsBlockStatus {
+    coeff1: i32,
+    coeff2: i32,
+    delta: i32,
+    sample1: i32,
+    sample2: i32,
+}
+
+impl AdpcmMsBlockStatus {
+    fn read_mono_preample<B: ReadBytes>(
```
pdeljanov, line 71: `read_mono_preamble`?

symphonia-codec-adpcm/src/codec_ms.rs
```diff
@@ -0,0 +1,167 @@
...
+            coeff1: params.coeffs1[block_predictor],
+            coeff2: params.coeffs2[block_predictor],
+            delta: i16_to_i32!(stream.read_u16()?),
+            sample1: i16_to_i32!(stream.read_u16()?),
+            sample2: i16_to_i32!(stream.read_u16()?),
+        };
+        Ok(status)
+    }
+
+    fn read_stereo_preample<B: ReadBytes>(
```
pdeljanov, line 87: `read_stereo_preamble`?

symphonia-codec-adpcm/src/common.rs
```diff
@@ -0,0 +1,29 @@
...
+impl Nibble {
+    pub fn get_nibble(&self, byte: u8) -> u8 {
+        match self {
+            Nibble::Upper => byte >> 4,
+            Nibble::Lower => byte & 0x0F,
+        }
+    }
+}
+
+macro_rules! i16_to_i32 {
```
pdeljanov, line 16: Should this be called `u16_to_i32` instead?

symphonia-codec-adpcm/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        let frames = match params.max_frames_per_packet {
+            Some(frames) => frames,
+            _ => return unsupported_error("adpcm: maximum frames per packet is required"),
+        };
+
+        if params.frames_per_block.is_none() {
+            return unsupported_error("adpcm: frames per block is required");
+        }
```
pdeljanov, line 135: Check that the value is > 0 since it's used for a division in the decode step.

symphonia-format-wav/src/chunks.rs
```diff
@@ -506,15 +574,20 @@ impl ParseChunk for WaveFormatChunk {
...
         const WAVE_FORMAT_ALAW: u16 = 0x0006;
         const WAVE_FORMAT_MULAW: u16 = 0x0007;
+        const WAVE_FORMAT_ADPCM_IMA: u16 = 0x0011;
         const WAVE_FORMAT_EXTENSIBLE: u16 = 0xfffe;
 
         let format_data = match format {
             // The PCM Wave Format
             WAVE_FORMAT_PCM => Self::read_pcm_fmt(reader, bits_per_sample, n_channels, len),
+            // The Microsoft ADPCM Format
+            WAVE_FORMAT_ADPCM => {
```
pdeljanov, line 588: A similar case probably needs to be implemented for the extensible WAVE format block. See the implementation of `read_ext_fmt`.

pdeljanov (commented):
Hey, this is really great! Thanks for implementing these decoders!

There are some minor changes that need to be made, and a couple other things that I think need to be double checked, but overall it looks good to me.

A couple general items:

1. Please add yourself to the CONTRIBUTORS file
2. It's not very obvious, but you need to run a nightly version of `cargo fmt` for the code to be formatted correctly.

In terms of testing, have you ran your test files through `symphonia-check`? 

I transcoded a couple files to ADPCM and checked them, but that's a very small test.

Thanks again, and hope to get this merged soon!

symphonia-format-wav/src/chunks.rs
geckoxx: I do not have any test files for it. So it is not implemented yet.

geckoxx:
> Just FYI, FWIW there's also an ADPCM implementation here: https://github.com/ruffle-rs/ruffle/blob/master/core/src/backend/audio/decoders/adpcm.rs
> 
> It might be too simplistic (it has to support only what Flash Player did), but it could still be interesting for cross-referencing.

I will look into it. Seems to be the IMA Codec.

geckoxx:
I tested my 2868 files (all with the Microsoft ADPCM Codec) and all passed.
I tested also the 16 ADPCM files from [this side](http://mauvecloud.net/sounds/index.html) and all passed.

symphonia-codec-adpcm/src/codec_ms.rs
pdeljanov: Oh that's a helpful document, thanks for that. It clarifies things.

So the extra data always provide the standard built-in coefficients, but they can extend it with more. In practice though, it's never extended.

pdeljanov (approved):

pdeljanov:
@geckoxx, thanks for making those changes.

Things look good to me, and your test results are very promising. Once you are ready we can merge the decoder and let it sit in the main branch for a bit.

I'm sure some people will experiment with it, and perform further testing.

pdeljanov:
Thanks for your work, @geckoxx! Merged!

PR #167, opened 2022-12-06
play: Add resampling for Windows.

erikas-taroza:
Fixes: https://github.com/pdeljanov/Symphonia/issues/43

The symphonia-play example will now use [libsamplerate](https://github.com/Prior99/rust-samplerate) to resample according to the default output config from Windows.


pdeljanov:
Hey @erikas-taroza,

Thanks for addressing this, I'm sure Windows users will appreciate it!

`libsamplerate` is a great quality library, but did you consider [`rubato`](https://docs.rs/rubato/latest/rubato/)?

Rubato is a native Rust resampling library. I think it may be worthwhile to consider it because people often just copy `symphonia-play`'s code, and having a Rust-based resampler would be more convenient for them. As a bonus, it'll likely build faster.

erikas-taroza:
Yes, the downside of ``libsamplerate`` is building (you need cmake installed on Windows). The reason I went for ``libsamplerate`` is because it is very easy to use. You give it a ``Vec<f32>`` and you get a ``Vec<f32>`` back. I am a beginner in Rust and audio stuff, so that was nice.

There is also a [fork](https://github.com/3c1u/rust-samplerate) of the ``libsamplerate`` repo that is pure Rust.

I will give ``rubato`` another go and compare the quality of the results.

erikas-taroza:
Hello @pdeljanov

I have implemented the resampling using `rubato`. Quality sounds good to me.

Turns out the reason I had so much trouble previously was that I needed to give it samples in planar format.

pdeljanov:
Hi @erikas-taroza,

Thanks for the quick turnaround on this! I'll spin up a Windows VM to try it.

I think with a couple minor Symphonia audio API additions we can get rid of all the allocations being done in the resampler code path. Do you mind if I push to your PR?

erikas-taroza:
Feel free to modify anything as needed :)

pdeljanov:
Hey @erikas-taroza,

I pushed the improvements in commit `1ab14583`.

Do you mind giving this a test? VM audio is not great unfortunately.

These are the changes I've made:
1. I pre-allocated all the buffers in `Resampler::new` since they should never need to change in capacity.
2. I used Symphonia's `AudioBufferRef::convert` function to perform the conversion from the decoded `AudioBufferRef` to an `AudioBuffer<f32>` for Rubato. This is convenient because `AudioBuffer` can then provide a slice-of-slices of all audio planes to Rubato.
3. I perform a final sample format conversion and interleaving in `Resampler::resample` after resampling and return a slice to the interleaved audio. This allows `AudioOutput::write` to be greatly simplified.

---

This code is okay and could be merged as-is, but there are some further improvements that could be made. If you are interested, these are:

1. The output's `SampleBuffer` is not used if resampling. It's therefore a waste to allocate it. Instead of making `Resampler` an `Option`, it could be an enum with two variants: one for resampling (with `Resampler`), and one when not resampling (with a the `SampleBuffer`).
2. If the `AudioBufferRef` passed to `Resample::resample` is a `f32` variant, then `convert` just copies the samples into the new buffer verbatim. This is wasteful since the samples from the original buffer could be used instead.

erikas-taroza:
Hello. The improvements look good :)

Just tested this and it works except for one file. It throws this error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Wrong number of frames 47 inn input channel 0, expected 1152', symphonia-play\src\resampler.rs:59:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I fixed this in my commit before your changes but it seems they got removed.

This is the link to the file. I tested the 2MB version.
https://file-examples.com/index.php/sample-audio-files/sample-mp3-download/

pdeljanov:
Oh I see!

Rubato requires the same number of samples to be provided each time. This is a problem because some decoders return a variable number of samples (to a maximum of the `duration` passed into `Resampler::new`).

Sadly this means we'll need to collect input samples in the resampler, and then only ask Rubato to resample a fixed size block of samples once we collect enough.

When there aren't enough samples, the resampler can "ask" for more by simply returning an empty slice. The only edge case is when the file ends and there are no more samples to be had. In this case the resampler will also need a `flush` function that pads the collected amount of samples with 0 up-to the block size and then resamples. This `flush` function can be called to acquire the last block of samples in `AudioOutput::flush` which can then write it to the ring buffer.

Do you want to have a go at this?

erikas-taroza:
Sure, I can fix it.

erikas-taroza:
@pdeljanov I have fixed the issue. I used `self.input.fill()` to fill the rest of the planes with 0s. Let me know if there is a better solution because you have a good understanding of this compared to me.

pdeljanov:
Ah okay, I think there was some confusion. I may have explained poorly.

Basically, a decoder can (and will) return a differing number of audio frames per packet. The duration the resampler is initialized with is the maximum number of frames that a decoder can return. Rubato requires a fixed number of input frames to operate and we configure this to be the aforementioned maximum. Therefore, there are cases where we must decode more than one packet to collect enough audio frames before we can resample.

In your solution, you padded the input with zero, but this would distort the audio because in most cases we can get more samples by decoding another packet. To put it another way, you added a random silent section into the audio track where there shouldn't be any.

What you implemented is correct for when we've consumed all the packets, but is not correct for the regular case where we do have more packets.

The audible glitches probably are not noticeable with testing because only the Vorbis codec (not commonly used) varies the number of audio frames per packet. In all other cases, the cases where we would get less samples than expected would be at the start or the end of a track which is usually silent anyways.

---

In terms of solutions, I don't think it's going to be very easy to use a Symphonia `AudioBuffer`s to collect the samples. Likely we will need to use to a `Vec<Vec<f32>>` on the input side to collect the samples.

Let me know if you want me to take the wheel, I understand that this may be getting in the weeds.

erikas-taroza:
Ah I see. I never tested with Vorbis so the solution did work fine for things like MP3. I will give this another go.

Ideally, I don't want the user to be calling a `flush` method for the last packet. Is this something that is unavoidable?

pdeljanov:
> Ah I see. I never tested with Vorbis so the solution did work fine for things like MP3. I will give this another go.
> 
> Ideally, I don't want the user to be calling a `flush` method for the last packet. Is this something that is unavoidable?

Yeah, it's unavoidable.

This is because `symphonia-play` uses a push-based architecture. The decode loop pushes audio buffers to the audio output. The decode loop is throttled by the block writing into the audio output ring buffer. When there are no more packets, the loop will exit regardless of whether or not the resampler is waiting for more samples or not.

After the decode loop exits, `symphonia-play` calls `AudioOutput::flush` which waits for the audio output to play the remaining samples. Otherwise `symphonia-play` would exit before the entire track is played. In this routine you'll also have to flush the resampler (if it's being used) too. When flushing the resampler you'd pad the buffer with zeros (like you did now), resample the buffer, and then write the last samples to the ring buffer.

A really good implementation would also chop off the extra zeros after resampling. The number of extra zeros after resampling would be (amount of added padding zeros before resampling) * (new sample rate) / (old sample rate). You can leave this out if you want. It's not particularly important to the demo app.

erikas-taroza:
Doesn't AudioOutput::flush for the cpal implementation just pause the stream?

pdeljanov:
> Doesn't AudioOutput::flush for the cpal implementation just pause the stream?

Apologies for the confusion. I was thinking of the PulseAudio implementation.

Ideally `AudioOutput::flush` does as I described. However, `cpal` doesn't appear to have a way to flush a stream the same way PulseAudio does, but it looked like the pause semantics were as close as we could get some of the times, see [here](https://github.com/RustAudio/cpal/issues/284).

You are right though, it's likely on some platforms we are probably dropping the output. A more correct implementation may be to wait until our ring buffer is emptied.

erikas-taroza:
Ok. I will keep all of this in mind as I work on this.

roderickvd:
Does Rubato’s FftFixedOut not allow for a varying number of input samples? https://docs.rs/rubato/latest/rubato/struct.FftFixedOut.html

pdeljanov:
> Does Rubato’s FftFixedOut not allow for a varying number of input samples? https://docs.rs/rubato/latest/rubato/struct.FftFixedOut.html

To my understanding[^1], the "Fixed" part reads more like "the parameter you choose":
- `FftFixedOut` means you choose how many frames you want at the output for each `process` call, but Rubato decides how many you need to input
- `FftFixedIn` means you choose how many frames you want to input for each `process` called, but Rubato decides how many will be output

In both these cases you need to provide Rubato with a certain number of input frames, and that number is constant after initialization. Therefore, since it's not possible to control how many frames are output by a decoder, and that number is variable, then you will always need to collect frames before invoking Rubato.

[^1]: Based off the example provided by Rubato [here](https://github.com/HEnquist/rubato/blob/master/examples/fftfixedout64.rs)

roderickvd:
What you write makes sense. I got the idea because something somewhat related came up in CamillaDSP recently: https://github.com/HEnquist/camilladsp/commit/db8b696665284b61ab73251a878227cd8548e4e8

Instead of flushing with a zero-initialised slice, it sends `None` through an undocumented function of the resampler to flush without clipping.

Hope I'm not causing a ruckus here. Admittedly I'm a bit on thin ice here, trying to tie together ends if there are any.

erikas-taroza:
@pdeljanov I refactored the code. Also tested with Vorbis. I doubt its completely optimized though.

pdeljanov:
> What you write makes sense. I got the idea because something somewhat related came up in CamillaDSP recently: [HEnquist/camilladsp@db8b696](https://github.com/HEnquist/camilladsp/commit/db8b696665284b61ab73251a878227cd8548e4e8)
> 
> Instead of flushing with a zero-initialised slice, it sends `None` through an undocumented function of the resampler to flush without clipping.
> 
> Hope I'm not causing a ruckus here. Admittedly I'm a bit on thin ice here, trying to tie together ends if there are any.

No worries.

I think Rubato just has a low-level interface that we just need to work around. Unfortunately, there's a bit of an impedance mismatch with Symphonia that makes it cumbersome. I'll be looking to rework the `core::audio` APIs in 0.6 to improve this situation in general.

Stay tuned because I'll likely be creating an issue to request feedback and comments before committing to the new API.

> @pdeljanov I refactored the code. Also tested with Vorbis. I doubt its completely optimized though.

Thanks! I haven't been able to dig into this yet, but I will try in the coming days.

pdeljanov:
Hi, thanks for waiting!

This turned out to be quite a bit more difficult than I anticipated.

While correct, the original code performed a lot of allocations which is generally not great when dealing with audio. So I worked to remove all allocations in the code path. However, providing a slice of exact sized slices to the resampler proved difficult. I ended up settling on using `arrayvec` to create a stack allocated vector to push the slices to.

I'm not particularly happy with how complex this ended up being, but the mismatch in interfaces was difficult to deal with.

I also needed to remove the code you added to flush samples. Unfortunately, it behaves incorrectly when there are underruns (gets detected as "the end"). I think it's best we leave this part out for now and only focus on resampling in this PR. Realistically, I think a proper solution would need to be implemented in `cpal`.

I tested these changes by forcing a build on my Mac to resample. It seems to work. If it works for Windows, then I think we can merge this unless you see any issues with it.

erikas-taroza:
Hello.

I will test this within the next day :)

erikas-taroza:
@pdeljanov Hello. I have tested the new code. It resamples on Windows. Time to merge :)

pdeljanov:
@erikas-taroza, merged! Thanks for your contribution and help!

PR #180, opened 2023-01-27
Handle incorrect channel mask in WAV files

jesnor:
This PR contains a work around for some WAV files that have an incorrect channel mask value. For example VLC can still play them.

### DIFF
--- a/symphonia-format-wav/src/chunks.rs
+++ b/symphonia-format-wav/src/chunks.rs
@@ -41,6 +41,10 @@ impl ParseChunkTag for NullChunks {
 fn fix_channel_mask(mut channel_mask: u32, n_channels: u16) -> u32 {
     let channel_diff = n_channels as i32 - channel_mask.count_ones() as i32;
 
+    if channel_diff != 0 {
+        info!("Channel mask not set correctly, channel positions may be incorrect!");
+    }
+
     // Check that the number of ones in the channel mask match the number of channels.
     if channel_diff > 0 {
         // Too few ones in mask so add extra ones above the most significant one
--- a/symphonia-format-wav/src/chunks.rs
+++ b/symphonia-format-wav/src/chunks.rs
@@ -38,6 +38,38 @@ impl ParseChunkTag for NullChunks {
     }
 }
 
+fn fix_channel_mask(mut channel_mask: u32, n_channels: u16) -> u32 {
+    let channel_diff = n_channels as i32 - channel_mask.count_ones() as i32;
+
+    // Check that the number of ones in the channel mask match the number of channels.
+    if channel_diff > 0 {
+        // Too few ones in mask so add extra ones above the most significant one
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hi @jesnor,

I've considered this workaround before since I've seen a single file with this issue. I didn't implement it because it didn't seem common and also because the channel mapping will be very wrong if > 2 channels.

Do you think we should limit this workaround if only `n_channels` is 1 or 2 (if 0, the workaround shouldn't be applied)? The mapping wouldn't be wrong in that case. Also, we'll definitely need a warning log if the workaround is applied.

Do you also happen to know if the WAV files were produced by a particular app or encoder?

Thanks!

jesnor:
The file that failed for me had 9 channels, created with Audacity I think. Interestingly similar file with 8 channels had correct channel mask.

Since VLC can play the file, I think the patch is reasonable. It's better to be able to decode something rather than nothing, although the channel mask might be weird.

pdeljanov:
Found [this](https://learn.microsoft.com/en-us/previous-versions/windows/hardware/design/dn653308(v=vs.85)?redirectedfrom=MSDN#details-about-dwchannelmask) documentation from Microsoft. Looks like `0` just means the channels are not assigned a "position", or are ambiguous and the decoder can device.

We don't have a way to express un-positioned channels with Symphonia's current API, so your solution is valid for all channels then. Only other alternative would be to assume certain channel layouts for each number (i.e., 2.1 for 3, etc.).

Just going to leave a couple small comments and then I'll merge the change once fixed.

symphonia-format-wav/src/chunks.rs
```diff
@@ -394,7 +394,12 @@ impl WaveFormatChunk {
             );
         }
 
-        let channel_mask = reader.read_u32()?;
+        let mut channel_mask = reader.read_u32()?;
+
+        // Graceful handling of "empty" channel mask value. Use the first n_channels channels.
+        if channel_mask == 0 {
```
pdeljanov, line 400: If `n_channels` is also 0 then this should be executed because of the next comment.

symphonia-format-wav/src/chunks.rs
```diff
@@ -394,7 +394,12 @@ impl WaveFormatChunk {
...
+            channel_mask = (1 << n_channels) - 1;
```
pdeljanov, line 401: Add an `info!` log here stating "channel mask is empty, channel positions may be incorrect" or something similar.

pdeljanov (changes requested):

symphonia-format-wav/src/chunks.rs
jesnor: Pushes a better version that should handle all cases (too many/few bits in mask) according to the document you linked.

pdeljanov (approved):

pdeljanov:
Thanks! This is great!

PR #182, opened 2023-01-30
 Add a fuzzer for the new, much more complex FFT implementation 

Shnatsel:


Shnatsel:
No issues after 10 million executions. 

I've run it without overflow checks and debug assertions, however. I was only interested in release-mode panics and memory safety issues.

pdeljanov:
I've reworked the `rustfft` integration in `master` so that it's only used internally by the `Imdct`. This is because using it for `Fft` forces a SemVer breaking change. In `v0.6` I'll switch `Fft` over.

Aside from that, don't you think fuzzing `rustfft` should be done in the `rustfft` repo?

Shnatsel:
I'm working on fuzzing in the RustFFT repo, yes: https://github.com/ejmahler/RustFFT/issues/110

But since you said that Symphonia uses RustFFT in specific and limited ways, I thought it might be a good idea to fuzz that codepath first, which should be simpler than the general case.

The RustFFT APIs are more complicated and I still haven't built a working fuzzer in the upstream repo. I could move this code there, if you wish.

PR #186, opened 2023-02-10
Fix up MusicBrainz reading

StratusFearMe21:
This PR fixes up how Symphonia reads MusicBrainz data out of ID3v2 tags. It fixes the TXXX `StandardTags` based on the tags that [MusicBrainz Picard maps to](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html). And it adds UFID frame support so decoding the MusicBrainz recording ID is possible


StratusFearMe21:
Hmmm, seems like `rustfmt` changed some `else` clauses in the code. If that's a style error, I could also add a `rustfmt.toml` to this PR

StratusFearMe21:
Not sure how to request a review of this PR in this repo, but could someone do that? I'd love to get this merged in the main crate

pdeljanov:
I think a lot of this is likely fixed/improved already in the `dev-0.6` branch. Though, I'm still iterating on the metadata interfaces. Some of the MusicBrainz tag mappings may be missing, however.

I'm not keen on picking up all these changes in `master` which is only reserved for very small fixes now, however, if you find any missing mappings in `dev-0.6` I'd be happy to have them.

PR #188, opened 2023-02-13
Switch to webm-iterable for parsing raw file data

austinleroy:
Hello! I saw the post for Symphonia on Reddit a little while back and it immediately piqued my interest.  I love writing code in Rust, and I've actually done a fair amount of work around parsing and writing EBML (basis for mkv/webm). I decided to look into the project to see if there was anything I could do to help on the Matroska front.

I put together this PR in an attempt to:
- Reduce code in the project
- Make the mkv demuxer logic operate at a higher level (deal with metadata packets rather than need to know binary encodings)
- Improve stability/compatibility
- Improve demuxing speed

Overall, I'm satisfied with the outcome.  This PR does remove over 1000 lines of code (822 additions and 2078 deletions), it delegates raw EBML packet parsing to the webm-iterable library (some bias here since I wrote it), and the resulting `symphonia-play` binary can play all of the [matroska test files](https://github.com/ietf-wg-cellar/matroska-test-files).  I unfortunately wasn't able to improve the demuxing speed like I had hoped, but the speed also isn't any worse, so a neutral result there.

Hopefully you think this is a helpful change!  I'm more than happy to talk through any questions or feedback.

> **Some notes on the [matroska test files](https://github.com/ietf-wg-cellar/matroska-test-files)**
> This branch plays them all perfectly except for file 7 (the damaged/corrupted file).  It plays, but not smoothly - it skips around a little more than `mpv` does.  Just a note that there could still be improvement there.
> As far as the previous version of the mkv demuxer - it failed to play test file 3 entirely, and test file 7 ended after 1 second of playback.  The other test files were fine, although some other mkv/mka files I found in the wild were also unplayable on the old version (the new version plays these files fine).  I've uploaded 2 examples in another branch in my forked repo: https://github.com/austinleroy/Symphonia/tree/AdditionalTestFiles/symphonia-play/mkvtestfiles

> **Some notes on my performance testing**
>  I used `hyperfine` and the `--decode-only` flag like was described elsewhere in this repo.  All of my runs seemed to bounce between the new version and the old version being "1.00 +- 0.02 times faster" than the other.  I took that as an indication that performance was identical, although it would be good to confirm this on other machines.


pdeljanov:
Matroska was rewritten as part of the `dev-0.6` branch. My preference is to keep dependencies on external crates to a minimum. However, anyone may write or publish a Symphonia-compatible demuxer or decoder crate. It'll just need to be manually registered into Symphonia by the developer using it.

PR #191, opened 2023-03-01
remove executable bit on *.rs files

tenzap:


pdeljanov:
Thanks, nothing wrong with it, just short on time. I hope these stray executable bits didn't cause you any problems.

PR #193, opened 2023-03-06
setup of riff module, support for PCM aiff

dedobbin:
I set up a RIFF module to support AIFF files. It's heavily based on the wav module. 
As discussed, in a future release the WAVE logic should be moved in to this RIFF module, since both WAVE and AIFF are contained in a RIFF file


symphonia-format-riff/src/chunks.rs
```diff
@@ -0,0 +1,299 @@
...
+        // Strictly, this must either be 8 or 16 bits, but there is no reason why 24 and 32 bits
+        // can't be supported. Since these files do exist, allow for 8/16/24/32-bit samples, but
+        // error if not a multiple of 8 or greater than 32-bits.
+        // 
+        // It is possible though for AIFF to have a sample size not divisible by 8.
+        // Data is left justified, with the remaining bits zeroed. Currently not supported.
+        //
+        // Select the appropriate codec using bits per sample. Samples are always interleaved and
+        // little-endian encoded for the PCM format.
+        let codec = match bits_per_sample {
```
pdeljanov, line 185: As mentioned, setting `bits_per_coded_sample` and `bits_per_sample` correctly in the `CodecParameters` will make `PcmDecoder` properly read non-multiple-of-8 samples. The codec type will need to be round up to the nearest multiple, though (e.g., 17 bits = `CODEC_TYPE_S24_BE`).

symphonia-format-riff/src/chunks.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+impl ParseChunk for CommonChunk {
+    fn parse<B: ReadBytes>(reader: &mut B, _tag: [u8; 4], _: u32) -> Result<CommonChunk> {
+        let n_channels = reader.read_be_i16()?;
+        let n_sample_frames = reader.read_be_u32()?;
+        let sample_size = reader.read_be_i16()?;
+        
+        let mut sample_rate: [u8; 10] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        let _res = reader.read_buf(sample_rate.as_mut());
```
pdeljanov, line 226: I don't think the `Result` should be ignored here. At a minimum, you need `?` so that if an IO error occurs the `parse` function will return the error.

Aside from that, why not `read_buf_exact`? The only condition where it would read less is if you reach the end of the file.

symphonia-format-riff/src/chunks.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+}
+
+impl ParseChunk for CommonChunk {
+    fn parse<B: ReadBytes>(reader: &mut B, _tag: [u8; 4], _: u32) -> Result<CommonChunk> {
+        let n_channels = reader.read_be_i16()?;
+        let n_sample_frames = reader.read_be_u32()?;
+        let sample_size = reader.read_be_i16()?;
+        
+        let mut sample_rate: [u8; 10] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
pdeljanov, line 225: Short hand way:
```suggestion
        let mut sample_rate: [u8; 10] = [0; 10];
```

symphonia-format-riff/src/chunks.rs
```diff
@@ -0,0 +1,299 @@
...
+        let _res = reader.read_buf(sample_rate.as_mut());
+        let sample_rate =  Extended::from_be_bytes(sample_rate);
+        let sample_rate = sample_rate.to_f64() as u32;
```
pdeljanov, line 228: Is a 0 sample rate an error?

symphonia-format-riff/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+/// Aiff is actually a RIFF stream, with a "FORM" ASCII stream marker.
+const AIFF_STREAM_MARKER: [u8; 4] = *b"FORM";
+
+/// A possible RIFF form is "aiff".
+const AIFF_RIFF_FORM: [u8; 4] = *b"AIFF";
+/// A possible RIFF form is "aifc", using compressed data.
+const AIFC_RIFF_FORM: [u8; 4] = *b"AIFC";
+
+/// The maximum number of frames that will be in a packet.
+/// Took this from symphonia-format-wav, but I don't know if it's correct.
```
pdeljanov, line 43: It's arbitrary since there are no packets in WAVE. 1152 is the same as MP3.

symphonia/src/lib.rs
```diff
@@ -161,6 +161,8 @@ pub mod default {
         pub use symphonia_format_mkv::MkvReader;
         #[cfg(feature = "ogg")]
         pub use symphonia_format_ogg::OggReader;
+        #[cfg(feature = "riff")]
+        pub use symphonia_format_riff::RiffReader;
```
pdeljanov, line 165: I think you are still experimenting with this. But presumably this should be an `aiff` feature, enabling the `AiffReader`.

pdeljanov:
You'll need a nightly toolchain to properly format the code! Once you install the nightly toolchain you can run `rustfmt` with `cargo +nightly fmt`.

symphonia-format-riff/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "symphonia-format-riff"
+version = "0.5.2"
+description = "Pure Rust RIFF demuxer (a part of project Symphonia)."
+homepage = "https://github.com/pdeljanov/Symphonia"
+repository = "https://github.com/pdeljanov/Symphonia"
+authors = ["Philip Deljanov <philip.deljanov@gmail.com>"]
```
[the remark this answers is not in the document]
dedobbin: Ah, done!

symphonia-format-riff/src/chunks.rs
dedobbin: I tried this and it does allow for playback, but the audio is clipping. I manually compared the values in the binary and what Symphonia was reading as samples, and it seemed to be multiplied by 16, but i haven't spent that much time digging into it

The CodecParameter value was properly set to 12 though, so something somewhere else is misbehaving.

symphonia-format-riff/src/chunks.rs
dedobbin: Ah you are totally right. I went with the minimal solution because i think in the case of a failure of reading the sample rate, there is not much recovery.

Definitely an argument to be made that the application should not stop on a broken file/read, but it's not an uncommon pattern in the library so i think it fits.

I can make it more dynamic though if you want, let me know!

symphonia-format-riff/src/chunks.rs
pdeljanov: Got it, it could be the decoder then.

symphonia-format-riff/src/chunks.rs
pdeljanov: I'd probably assume a sample rate of 0 is an error. At least, it's unplayable since samples would never be consumed.

symphonia/src/lib.rs
pdeljanov: Sorry, I'm a bit confused. I think there are 3 ways forward:

1) Have a `symphonia-format-aiff` crate with a single module `AiffReader`. Keep `WaveReader` in it's own `symphonia-format-wave` crate. Other RIFF based formats would also have their own crates and readers.
2) Have generic RIFF crate `symphonia-format-riff` with a single `RiffReader` that can read AIFF, WAVE, etc.
3) Have a generic RIFF crate `symphonia-format-riff` with multiple readers: `AiffReader`, `WaveReader`, etc.

I got the impression that we narrowed our options between 2 and 3, but it seems you landed on 1 in the current iteration. Have I missed anything?

dedobbin: As discussed, 3 it is!

pdeljanov:
Thanks for your work on this!

PR #194, opened 2023-03-07
Support multichannel WAV formats

jesnor:
Problems with some WAV file again :) This is using the "old" WAVEFORMAT file format with multichannel (>2). This fix just maps to the first n channels. I don't know if you want to accept it, I'm not sure if it's allowed in the WAV standard, but on the other hand maybe it's better to be able to use the file rather than giving an error. It can be played in VLC for example and the channel mapping is 3F2M2R/LFE, which I believe is the same as setting the first n bits of channel mask.

### DIFF
--- a/symphonia-format-wav/src/chunks.rs
+++ b/symphonia-format-wav/src/chunks.rs
@@ -74,6 +74,27 @@ fn test_fix_channel_mask() {
     assert_eq!(fix_channel_mask(0xFFFFFFFF, 8), 0b11111111);
 }
 
+fn try_channel_count_to_mask(count: u16) -> Result<Channels> {
+    (1..=32)
+        .contains(&count)
+        .then(|| Channels::from_bits(((1u64 << count) - 1) as u32))
+        .flatten()
+        .ok_or(Error::DecodeError("wav: invalid channel count"))
+}
+
+#[test]
+fn test_try_channel_count_to_mask() {
+    assert!(try_channel_count_to_mask(0).is_err());
+
+    for i in 1..27 {
+        assert!(try_channel_count_to_mask(i).is_ok());
+    }
+
+    for i in 27..u16::MAX {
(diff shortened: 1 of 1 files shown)


symphonia-format-wav/src/chunks.rs
```diff
@@ -305,13 +305,8 @@ impl WaveFormatChunk {
...
         };
 
-        // The PCM format only supports 1 or 2 channels, for mono and stereo channel layouts,
-        // respectively.
-        let channels = match n_channels {
-            1 => Channels::FRONT_LEFT,
-            2 => Channels::FRONT_LEFT | Channels::FRONT_RIGHT,
-            _ => return decode_error("wav: channel layout is not stereo or mono for fmt_pcm"),
-        };
+        let channels = Channels::from_bits((1 << n_channels) - 1)
```
pdeljanov, line 308: This will panic if `n_channels` is > 31 because the `Channels` bit-mask is 32 bits and shifts > 31 is undefined behaviour. So I think we'll need a check to reject files with too many channels or zero channels, and to ensure that the shift logic works for all cases that do pass the check.

It may make sense to even add a `try_new_from_count` (or some other name) function to `Channels` to centralize part of this logic (0 is valid for `Channels` but not for WAVE).

pdeljanov (changes requested):
Thanks for this!

I think mirroring the behaviour of VLC or FFmpeg here is reasonable. However, there are some potential issues that need to be taken care of to prevent panics.

symphonia-format-wav/src/chunks.rs
jesnor: Added try_channel_count_to_mask function with unit test

pdeljanov (approved):

pdeljanov:
Thanks, this looks good to me! Merged!

PR #199, opened 2023-04-04
core: Clean up Debug formatting of meta::Visual

zombiepigdragon:
This removes the often large and unreadable Debug formatting for the affected type (a raw `[u8]` dump for an embedded image file), and replaces the output for data with a manually-formatted type that only shows the data's size. An alternative would be using a wrapper type for `Box<u8>` that implements this manual debug format on its own, avoiding the possible maintenance burden this PR could have and being reusable for other blob-like data in Symphonia, but this was not chosen as it would likely become a breaking change for relatively little benefit.

### DIFF
--- a/symphonia-core/src/meta.rs
+++ b/symphonia-core/src/meta.rs
@@ -352,7 +352,7 @@ pub enum ColorMode {
 }
 
 /// A `Visual` is any 2 dimensional graphic.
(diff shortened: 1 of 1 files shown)


dedobbin:
I can imagine there are situations where you want to see the data, but you can always just print the .data member. So yeah makes sense and lgtm

PR #202, opened 2023-04-13
Track names

algesten:
Hi!

I've been testing out using the Demuxer directly using symphonia-core and symphonia-format-mkv. I'm quite used to dealing with demuxers and codecs, and using them works great!

For my use case, I'm decoding MKV- and MP4-files that have both audio and video, but I'm only interesting in extracting the audio. The containers also have multiple audio files, and they are differentiated by name. However the names are not currently surfaced in [Track](https://docs.rs/symphonia-core/0.5.2/symphonia_core/formats/struct.Track.html).

Reading the crates.io README for core and format-mkv, I see:

> Note: This crate should only be used if developing Project Symphonia decoders and demuxers. For other use cases please use the [symphonia](https://crates.io/crates/symphonia) crate.

I've been trying to read other PRs and Issues to understand what the stance is on using the lower level crates. Is it a supported use case or not? I totally get that it might not be.

As for the track names. Would you be open to a PR that starts doing that for MKV and MP4? You can see the start of my changes here.

### DIFF
--- a/symphonia-format-mkv/src/demuxer.rs
+++ b/symphonia-format-mkv/src/demuxer.rs
@@ -501,6 +501,8 @@ impl FormatReader for MkvReader {
                 id: track_id,
                 codec_params: codec_params.clone(),
                 language: track.language,
+                name: track.name,
+                ..Default::default()
             });
 
             states.insert(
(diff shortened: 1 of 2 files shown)


dedobbin:
I understand the need and it seems definitely valid, but i think it would be a better fit for metadata?
It would involve mapping the metadata to individual tracks though, which would probably broaden the scope of the changes.

PR #204, opened 2023-05-07
Feature/aifc

dedobbin:
This allows AiffReader to also read Aifc files. 
It works with all normal files from https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/Samples.html, with the exception of files with 12bit samples (as discussed, should be supported but for now it doesn't work yet because it causes distortion) and a file that uses sowt encoding.

Sowt support could be a cool feature for the future, but i think for now this is already usable.

Let me know what you think

**Edit**: regarding the test files, i only tested with the normal files, perhaps i should also check "perverse files" and "Other AIFF files"

**Edit again**: Tested it with the other files, some more missing compression formats (G722, G726, G728, GSM), and also some files using a specific block size and/or offset value in the sound chunk. It's currently not supported and i think i want to take a look at it, but seems out of scope for this PR

**Edit again again**: Turned out there was no real complexity in sowt encoding, i added it now.


AkiyukiOkayasu:
I am somewhat concerned that sowt is designed to handle only 16 bits.
I don't think this will be a problem for most use cases, though.
In fact, even ffmpeg fails when trying to convert a wav file to a 24bit little endian AIFF file.
```sh
ffmpeg -i input.wav -c:a pcm_s24le outputS24LE.wav
```

16bit little endian can be converted correctly.
```sh
ffmpeg -i input.wav -c:a pcm_s16le outputS16LE.wav
```

I could not find out if this is a specification of AIFF-C or a limitation of ffmpeg.
The following forum is not definitive, but it seems that libaiff is not limited to 16bit.
https://www.kvraudio.com/forum/viewtopic.php?t=328325

dedobbin:
Thank you for taking a look! 

"I am somewhat concerned that sowt is designed to handle only 16 bits."

I will try and dig up where i read about this being part of the SOWT encoding. It's not a problem to remove SOWT from this PR, it's rather out of scope anyway, but i couldn't find any files in the wild that disagreed with my implementation, so i figured why not.

Another option is to drop the 16bit constraint. It's not uncommon that some software exports audio files "off spec" to increase compatibility.

symphonia-format-riff/src/riff.rs
```diff
@@ -312,25 +316,21 @@ pub fn append_format_params(
...
-        FormatData::IeeeFloat(ieee) => {
-            codec_params.for_codec(ieee.codec).with_channels(ieee.channels);
-        }
-        FormatData::Extensible(ext) => {
-            codec_params
-                .for_codec(ext.codec)
-                .with_bits_per_coded_sample(u32::from(ext.bits_per_coded_sample))
-                .with_bits_per_sample(u32::from(ext.bits_per_sample))
-                .with_channels(ext.channels);
-        }
```
pdeljanov, line 327: I'm a bit confused about this. Wouldn't deleting some of this break the WAVE reading functionality?

symphonia-format-riff/src/aiff_chunks.rs
```diff
@@ -120,6 +233,64 @@ impl fmt::Display for CommonChunk {
...
+            n_channels,
+            n_sample_frames,
+            sample_size,
+            sample_rate,
+            format_data,
+            compression_type,
+            compression_name,
+        })
+    }
+}
```
pdeljanov, line 292: It looks like AIFC has some additional fields in the common chunk that are not present in AIFF. That means that the common chunk for AIFC is a superset of AIFF. So can you put the parsing of the common fields in `self.parse` and then just read the extra fields for AIFC after? This would reduce code duplication!

symphonia-format-riff/src/lib.rs
```diff
@@ -92,16 +97,11 @@ impl FormatReader for AiffReader {
...
-        }
-        else {
-            error!("riff form is not supported ({})", String::from_utf8_lossy(&riff_form));
-            return unsupported_error("aiff: riff form is not supported");
-        }
+        let riff_form = match riff_form {
+            AIFF_RIFF_FORM => AiffType::Aiff,
+            AIFC_RIFF_FORM => AiffType::Aifc,
+            _ => return unsupported_error("aiff: riff form is not supported"),
+        };
```
pdeljanov, line 104: You probably don't really need to introduce a new `enum` for this. Just use match the `riff_form` string like this below when you parse the common chunk.

symphonia-format-riff/src/aiff_chunks.rs
```diff
@@ -30,6 +35,10 @@ pub struct CommonChunk {
     pub sample_rate: u32,
     /// Extra data associated with the format block conditional upon the format tag.
     pub format_data: FormatData,
+    /// The compression ID, only exists for aifc
+    pub compression_type: [u8; 4],
+    /// The compression name, only exists for aifc
+    pub compression_name: String,
```
pdeljanov, line 41: I don't think you need  to store `compression_type` and `compression_name` in the `CommonChunk` struct because it's made redundant for both AIFF and AIFC by `format_data`. Then, you would need these magic strings.

pdeljanov (changes requested):

symphonia-format-riff/src/riff.rs
dedobbin: You are correct. I messed up merging my AIFC and AIFF branches, sloppy on my part, good catch. pushed the remaining code

symphonia-format-riff/src/aiff_chunks.rs
dedobbin: As discussed, this is desirable, but to do this "right", it probably involves the removing of the ParseChunk trait. I think that's a good idea and want to work on this, but i think it will balloon this PR. 

I propose for now, we uses this sub-optimal solution, and i can refactor it while i work on above mentioned task.

dedobbin: Al right, as discussed we will roll with this for now but i will add TWOS support

pdeljanov (approved):

PR #209, opened 2023-05-28
fix ogg seeking returning an out of bounds error for short files

### LINKED ISSUE #207: SeekError::OutOfRange when seeking to timestamp 0
The following code returns a `SeekError::OutOfRange` with [this audio file](https://github.com/pdeljanov/Symphonia/files/11582986/test.zip):

```rust
use std::{error::Error, fs::File};

use symphonia::core::{
	formats::{FormatOptions, SeekMode, SeekTo},
	io::{MediaSourceStream, MediaSourceStreamOptions},
	meta::MetadataOptions,
	probe::Hint,
};

fn main() -> Result<(), Box<dyn Error>> {
	let probe = symphonia::default::get_probe();
	let mss = MediaSourceStream::new(
		Box::new(File::open("test.ogg")?),
		MediaSourceStreamOptions::default(),
	);
	let mut format_reader = probe
		.format(
			&Hint::default(),
			mss,
			&FormatOptions::default(),
			&MetadataOptions::default(),
		)?
		.format;
	let track = format_reader.default_track().unwrap();
	format_reader.seek(
		SeekMode::Accurate,
		SeekTo::TimeStamp {
			ts: 0,
			track_id: track.id,
		},
	)?;

	Ok(())
}
```

The ogg does have audio, so I don't think this is the correct behavior. I also tried this with mp3, flac, and wav files with the same audio data, and none of them have this issue.

tesselode:
I have no idea if this is the best way to fix it, but it seems to make seeking work correctly for [this test file](https://github.com/pdeljanov/Symphonia/files/11585525/silence.zip) and hasn't broken anything else in my limited testing.

Fixes #207.

### DIFF
--- a/symphonia-format-ogg/src/demuxer.rs
+++ b/symphonia-format-ogg/src/demuxer.rs
@@ -123,14 +123,18 @@ impl OggReader {
 
             let mut start_byte_pos = self.phys_byte_range_start;
             let mut end_byte_pos = physical_end;
+            let mut first_loop = true;
 
             // Bisection method.
             loop {
-                // Find the middle of the upper and lower byte search range.
-                let mid_byte_pos = (start_byte_pos + end_byte_pos) / 2;
+                // Pick a byte to start searching from. For the first iteration,
+                // use the start byte. For each subsequent iteration, find the middle
+                // of the upper and lower byte search range.
+                let search_byte_pos =
+                    if first_loop { start_byte_pos } else { (start_byte_pos + end_byte_pos) / 2 };
 
-                // Seek to the middle of the byte range.
-                self.reader.seek(SeekFrom::Start(mid_byte_pos))?;
+                // Seek to the chosen search byte.
+                self.reader.seek(SeekFrom::Start(search_byte_pos))?;
 
                 // Read the next page.
                 match self.pages.next_page_for_serial(&mut self.reader, serial) {
@@ -142,19 +146,19 @@ impl OggReader {
                 let (start_ts, end_ts) = stream.inspect_page(&self.pages.page());
 
                 debug!(
-                    "seek: bisect step: page={{ start={}, end={} }} byte_range=[{}..{}], mid={}",
-                    start_ts, end_ts, start_byte_pos, end_byte_pos, mid_byte_pos,
+                    "seek: bisect step: page={{ start={}, end={} }} byte_range=[{}..{}], search={}",
+                    start_ts, end_ts, start_byte_pos, end_byte_pos, search_byte_pos,
                 );
 
                 if required_ts < start_ts {
                     // The required timestamp is less-than the timestamp of the first sample in
                     // page1. Update the upper bound and bisect again.
-                    end_byte_pos = mid_byte_pos;
+                    end_byte_pos = search_byte_pos;
                 }
                 else if required_ts > end_ts {
                     // The required timestamp is greater-than the timestamp of the final sample in
                     // the in page1. Update the lower bound and bisect again.
-                    start_byte_pos = mid_byte_pos;
+                    start_byte_pos = search_byte_pos;
                 }
                 else {
                     // The sample with the required timestamp is contained in page1. Return the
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks for attempting this!

This part of the seek code is using a binary search to roughly find the OGG page (a group of packets) where the desired packet is. A linear scan follows to find the exact packet.

It seems like the issue occurs because there are no more pages after the seek to the mid-point, so when it tries to find the next page after the seek there are none left.

Therefore, a possible simple solution could just be adding a condition after calculating the mid-point, but before the seek, that breaks out of the binary search:

```rust
if end_byte_pos - start_byte_pos <= 2 * OGG_PAGE_MAX_SIZE as u64 {
    break;
}
```

This same condition is used to terminate the binary search further down as well.

Disclaimer: I haven't really tested this code beyond your broken file, but it makes sense.

pdeljanov:
I'm not sure how familiar you are with the OGG container format, but the [documentation](https://xiph.org/ogg/) is pretty good at explaining the structure of the format.

> Is the point of that condition to check if there's at least 3 pages?

So, the thought process was that after the seek we need to have a page to sync the reader to (after all, we're just seeking to random byte locations in the file). One way to know if there is a page present is if `end - mid >= MAX_PAGE_SIZE`. Since a page can never exceed a maximum size, if the amount of data remaining is larger than that maximum size, there must be a page. That can then be rewritten as `end - start >= 2 * MAX_PAGE_SIZE` since `mid` is halfway between `end` and `start`.

Consider the case where there is only a single page and we seek to `t=0`. Before the seek, the reader is synced to the start of the first page. On seek, `start` is initialized to the byte position of the first byte of the first page, and `end` is initialized to one past the last byte position of the first page. Therefore, `mid` will point to a position in the middle of the first page. If we seek to `mid`, then there are no more pages to sync the reader to and the out-of-range error will be returned. Of course, this is wrong, we just shouldn't have performed that seek because there was no guarantee another page was present. With the change applied, since the condition is not met, the binary search would be immediately halted and the error avoided.

Though, upon further thought, both of these solutions only work if there is only a single logical stream (track) in the file. Multiple logical streams would complicate matters since their pages are interleaved with the track we're seeking.

> Also, how do you recommend testing changes to the decoder? Do you have a suite of audio files you test with?

To be frank, I don't really have many OGG files. The internet archive would be a decent place to find some. @Shnatsel also did a lot of OGG/Vorbis testing back in the day with a decently large corpus.

tesselode:
@pdeljanov regarding multiple tracks, `SeekTo` has a `track_id` field. Would it be reasonable to just try to seek to the timestamp in the specified track, and if the timestamp is out of bounds for that track, then return an `OutOfBounds` error? Or do you mean something else by the pages being interleaved with the tracks and I'm not understanding the issue?

pdeljanov:
Sorry I missed this message.

> @pdeljanov regarding multiple tracks, `SeekTo` has a `track_id` field. Would it be reasonable to just try to seek to the timestamp in the specified track, and if the timestamp is out of bounds for that track, then return an `OutOfBounds` error? 

Yeah, the intention of the `track_id` field is to tell the `FormatReader` to seek that track to the timestamp specified.

> Or do you mean something else by the pages being interleaved with the tracks and I'm not understanding the issue?

I did mean something else however. Basically the issue is the binary search could exclude the last page of a track if the mid-point was within that page (the reader tries to synchronize to the next page which doesn't exist). This resulted in an end-of-stream error.

My suggestion was to only continue the binary search if we know there is at least one page **for the track** between the mid-point and the begin or end (depending which way we're going) boundaries. I suggested calculating the difference between `end` and `start` and comparing it to the maximum possible page size to determine if there must be a page present. This works if there is one track, because if there is a page present, it **must** be for that one track. However, if there are multiple tracks this is not true.

pdeljanov:
Hi @tesselode, thanks for attempting this PR. I'm going to need to close it however since fixing this issue required a more comprehensive change (a1d97f02421de6e29f1dc4ca3c9735073b3bd141).

PR #211, opened 2023-06-02
Copy sample to buffer method

r4v3n6101:
Hello! I think it may be useful, because `cpal` crate, for example, can receive raw data in bytes. So I added method to copy into buffer of bytes.


r4v3n6101:
My bad, I see the RawSampleBuffer

PR #217, opened 2023-07-10
Add impl From<Duration> for Time and impl From<Time> for Duration

### LINKED ISSUE #216: time::Duration
We're using Symphonia in librespot as you know, and I was looking to moving most if not all time related things to Durations for consistency and ease of use. Anyway I noticed a lack of Duration support in your time units and calculations. And I was wondering if that was on purpose or not and if you'd be open to something like this?:

```rust
use std::{fmt, time::Duration};

...

impl From<Duration> for Time {
    fn from(duration: Duration) -> Self {
        let seconds = duration.as_secs_f64();
        Time::new(seconds.trunc() as u64, seconds.fract())
    }
}

impl TimeBase {
...
    pub fn calc_time_as_duration(&self, ts: TimeStamp) -> Duration {
        let time = self.calc_time(ts);
        Duration::from_secs_f64(time.seconds as f64 + time.frac)
    }
...
}
```

This would allow us to pass a Duration almost directly when seeking and get a duration for our packet time. If it's something that you're not interested in that's fine to, as it's pretty trivial to convert between Time and Duration ourselves.

JasonLG1979:
Closes #216

### DIFF
--- a/symphonia-core/src/units.rs
+++ b/symphonia-core/src/units.rs
@@ -72,7 +72,7 @@ impl Time {
     }
 
     pub fn into_duration(&self) -> StdDuration {
-        StdDuration::from_secs_f64(self.seconds as f64 + self.frac)
+        StdDuration::new(self.seconds, (1_000_000_000.0 * self.frac) as u32)
     }
 }
 
--- a/symphonia-core/src/units.rs
+++ b/symphonia-core/src/units.rs
@@ -7,7 +7,7 @@
 
 //! The `units` module provides definitions for common units.
 
-use std::fmt;
+use std::{fmt, time::Duration as StdDuration};
 
 /// A `TimeStamp` represents an instantenous instant in time since the start of a stream. One
 /// `TimeStamp` "tick" is equivalent to the stream's `TimeBase` in seconds.
@@ -70,6 +70,10 @@ impl Time {
 
         Some(Time { seconds, frac })
     }
+
+    pub fn into_duration(&self) -> StdDuration {
+        StdDuration::from_secs_f64(self.seconds as f64 + self.frac)
+    }
 }
 
 impl From<u8> for Time {
@@ -115,6 +119,13 @@ impl From<f64> for Time {
         else {
             Time::new(0, 0.0)
         }
+    }
+}
+
+impl From<StdDuration> for Time {
+    fn from(duration: StdDuration) -> Self {
+        let seconds = duration.as_secs_f64();
+        Time::new(seconds.trunc() as u64, seconds.fract())
     }
 }
 
@@ -231,6 +242,7 @@ impl fmt::Display for TimeBase {
 #[cfg(test)]
 mod tests {
     use super::{Time, TimeBase};
+    use std::time::Duration;
 
     #[test]
     fn verify_timebase() {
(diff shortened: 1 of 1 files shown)


symphonia-core/src/units.rs
```diff
@@ -70,6 +70,10 @@ impl Time {
 
         Some(Time { seconds, frac })
     }
+
+    pub fn into_duration(&self) -> StdDuration {
+        StdDuration::from_secs_f64(self.seconds as f64 + self.frac)
```
pdeljanov, line 75: A `f64` can only represent an unsigned integer up-to 2^52 without losing precision, so in extreme cases where seconds is very large, this conversion will be lossy.

Using `Duration::new` and passing the `u64` seconds component directly will preserve the accuracy in those extreme cases. To implement this, the fractional part must be converted to nanoseconds.

```suggestion
        StdDuration::new(self.seconds, (1_000_000_000.0 * self.frac) as u32)
```

symphonia-core/src/units.rs
```diff
@@ -118,6 +122,13 @@ impl From<f64> for Time {
     }
 }
 
+impl From<StdDuration> for Time {
+    fn from(duration: StdDuration) -> Self {
+        let seconds = duration.as_secs_f64();
+        Time::new(seconds.trunc() as u64, seconds.fract())
```
pdeljanov, line 128: Same idea about improving precision by removing the round-trip through a `f64`.

```suggestion
        Time::new(duration.as_secs(), f64::from(duration.subsec_nanos()) / 1_000_000_000.0)
```

pdeljanov (changes requested):
Thanks for this. Can you give these suggestions a try. I believe it should improve the accuracy in extreme cases.

symphonia-core/src/units.rs
```diff
@@ -7,7 +7,7 @@
 
 //! The `units` module provides definitions for common units.
 
-use std::fmt;
+use std::{fmt, time::Duration as StdDuration};
```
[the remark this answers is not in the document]
JasonLG1979: Done

symphonia-core/src/units.rs
JasonLG1979: Done.

symphonia-core/src/units.rs
JasonLG1979: Done.

pdeljanov:
Sorry for the possible churn, I had a last second thought about this. Do you think it'd be more idiomatic if instead of `into_duration()` we add `impl From<Time> for std::time::Duration {}`? Then one would just do `some_time.into()`.

JasonLG1979:
> Sorry for the possible churn, I had a last second thought about this. Do you think it'd be more idiomatic if instead of `into_duration()` we add `impl From<Time> for std::time::Duration {}`? Then one would just do `some_time.into()`.

It's all good. Whatever you think is best. I am by no means a judge on what is more idiomatic.

JasonLG1979:
There is some symmetry in impl'ing From for both that I can appreciate though. But I'm not really sure if it makes it more idiomatic or not?

JasonLG1979:
Done.

pdeljanov (approved):

pdeljanov:
I'm no expert either, but I do believe this very likely is the most idiomatic way. So let's go with it. Thanks for the making the changes.

PR #218, opened 2023-07-23
Feature/12 bit aiff

dedobbin:
Support for more sample sizes in aiff and wave


symphonia-codec-pcm/src/lib.rs
```diff
@@ -136,6 +157,25 @@ macro_rules! read_pcm_unsigned {
     };
 }
 
+macro_rules! read_pcm_unsigned_be {
+    ($buf:expr, $fmt:tt, $read:expr, $width:expr, $coded_width:expr) => {
+        // Get buffer of the correct sample format.
+        match $buf {
+            GenericAudioBuffer::$fmt(ref mut buf) => {
+                let sample_size = mem::size_of_val(&$read) * 8;
```
pdeljanov, line 165: I'm a bit confused by this because `$read` is an expression which will expand into something like `reader.reader.read_be_u32()`.

Wouldn't this erroneously consume a sample by mistake then?

dedobbin:
As we discussed, the current solution for working with BE encoded files that have a smaller coded width than sample width is correct. symphonia-check will fail on specific 12bit AIFF files, since ffmpeg treats them as 16bit.
According to the AIFF spec, this should not matter since the unused bits should be set to 0. In practice this turns out to not always be the case. It does not produce an audible difference though.

Because symphonia-codec-pcm is a seperate crate, not specific for RIFF files, we decided to ignore this edge case.

That being said, i had some difficulties finding more AIFF and WAVE test files. I did find some, but more testing could not hurt.

I did verify with symphonia-check that all test files i used when setting up the RIFF logic still work.

pdeljanov:
Hi @dedobbin,

Should we also apply the masking logic to LE codecs to be consistent with BE?

PR #220, opened 2023-08-15
Fixed MP4 atom length bug

jesnor:
Fixed a bug which causes u64 underflow when MP4 atom length is zero


pdeljanov:
Thank you! Nice find!

PR #224, opened 2023-08-30
skip invalid seek offset

davibe:
I came across (broken) content where the "+" would fail with overflow. The lib should not panic in this case.


k0nserv:
I wonder if this is fixing the underlying cause or just papering over something else. These are two `u64`s that are being added here, the resulting values is used as a byte offset for seeking. I don't think it's reasonable for this to overflow at all, even just 2^32 would be a 4Gb file. 

It seems likely that something about reading the `segment_pos` value goes wrong and results in an unreasonable value, which in turn causes the overflow.

dedobbin:
Could you share the file that caused this issue? I'm curious to see if it is indeed an underlying issue. Also i think if we roll with this, the println! should be a debug!  or warn!

k0nserv:
> Could you share the file that caused this issue? I'm curious to see if it is indeed an underlying issue. 

Unfortunately I can't. However, it's our own custom software stack, not any established MKV tool that generates the file, it's possible that there's an error in the our output.

I'll try to debug this at some point, but I have no timeline for that.

> Also i think if we roll with this, the println! should be a debug! or warn!

Agreed, I made it `info!` in our fork

PR #227, opened 2023-08-30
Make StandardTag{Key,Visual} implement Hash

ClementNerma:
This would come very handy when trying to get the value of a standard tag or visual multiple times, allowing for instance to build a `HashMap<StandardTagKey, Value>` and fetching from it instead of having to rely on a `&[Tag]` which requires iteration for each search.


pdeljanov:
Hi @ClementNerma,

Looks good to me. However, you'll need to use a nightly toolchain when running `rustfmt`. If you install a nightly toolchain you can simply run `cargo +nightly fmt`. The PR will not pass the formatting check otherwise.

ClementNerma:
Fixed.

pdeljanov:
Thanks!

PR #229, opened 2023-09-12
ogg: Decode opus packet durations

karlri:
It seems to me that this should decode opus packet durations correctly.
I implemented it because my buffering code seemed to hang trying to buffer up 1s of data forever. Perhaps it could be of interest?


karlri:
The PR is w

PR #231, opened 2023-09-13
ogg: Decode opus packet durations

karlri:
I implemented this because buffering some ogg web streams seemed to hang trying to buffer up 1s of data. Turns out it was opus and the durations of all opus packets were demuxed as 0. I think I implemented this correctly but don't take my word for it. Maybe it could be of use?


dedobbin:
Could you perhaps share some example of a files/stream where this faulty behavior occurs? I found some files that pass the altered code (https://www.dwsamplefiles.com/download-opus-sample-files/) but these all give "unsupported codec" so it's not of much use.

pdeljanov:
Hi @karlri, thanks for this! Apologies it took so long to get to.

PR #233, opened 2023-09-21
core: Use the PROBE_SEARCH_LIMIT when probing

perxjoh:
Change-Id: I691e457a66845e2a8e1e3b183a52e8e65b879895

### DIFF
--- a/symphonia-core/src/probe.rs
+++ b/symphonia-core/src/probe.rs
@@ -237,6 +237,10 @@ impl Probe {
 
             count += 1;
 
+            if count > Probe::PROBE_SEARCH_LIMIT {
+                break;
+            }
+
             if count % 4096 == 0 {
                 debug!(
                     "searching for format marker... {}+{} / {} bytes.",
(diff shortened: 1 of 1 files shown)


symphonia-core/src/probe.rs
```diff
@@ -291,8 +295,12 @@ impl Probe {
             }
         }
 
-        // Could not find any marker within the probe limit.
-        error!("reached probe limit of {} bytes.", Probe::PROBE_SEARCH_LIMIT);
+        if count < Probe::PROBE_SEARCH_LIMIT {
+            error!("probe reach EOF at {} bytes.", count);
```
dedobbin, line 299: I would personally do a `unsupported_error` here, because technically the problem is that the file is too short for Symphonia to make anything from it, rather than it not being recognized as valid format.

dedobbin:
If i understand correctly this just makes Symphonia more resilient against small (possible corrupted?) files. 
Get's a LGTM from me. Left a small comment/suggestion but it's more of a preference/style thing nothing mandatory

symphonia-core/src/probe.rs
pdeljanov: I think in both cases it is not actually possible to truly know if this file is a legitimate piece of media (in which case the file was either truncated or the media is prefixed with a lot of junk data). So, I'm okay with leaving it as-is.

pdeljanov:
Thanks for this!

I'm a bit surprised that the probe search limit was actually never used.

Could you please run cargo format with the nightly toolchain (the stable toolchain does not honour all the settings)?

pdeljanov:
Thank you!

Issue #171, opened 2022-12-26
Is it possible to replace `MediaSourceStream` with generic `M: MediaSource`?


compilerintrinsics:
Instead of taking in `MediaSourceStream`, is it possible to have decoders generically take in a `M: MediaSource` trait instead? (Or possibly `Box<dyn MediaSource>`, although I prefer the generic version because of fewer pointer indirections). The reason being, the underlying `MediaSource` implementation may already be buffered or is entirely in-memory, causing unnecessary copy into the buffer within `MediaSourceStream`.

The existing `MediaSourceStream` can be repurposed as `MediaSourceStream<M>`, a buffered wrapper type over `MediaSource` types, so that the user can decide for themselves whether they require buffering (pay only when needed).

pdeljanov:
I think this is a topic that could go very deep, but...

The short answer is that from my experience, I don't think there will be much of a gain in performance. Nothing measurable at least. Memory copies are extremely fast and optimized, and we are only copying tens of MB of memory. Removing a single branch from a `Decoder` hot-path would yield more benefits.

---

The long answer is that it's a complicated problem...

`MediaSourceStream` is basically  a `BufReader` with extended capabilities.

The motivation behind making `MediaSourceStream` take ownership of a `Box<dyn MediaSource>` is so the overhead of an indirect call is only paid when buffering new data. In this way, the cost of that indirect call is amortized over all the read calls on `MediaSourceStream` it takes to consume the newly buffered data.

If we made everything that used `MediaSourceStream` generic over `MediaSource` instead, that would solve the indirect call issue, but it would also just increase the complexity of everything else.

`MediaSourceStream` also provides the following invaluable capabilities:

1. The ability to grow the underlying buffer as needed
2. The ability to seek within the buffered data without seeking the underlying source
3. Optimized single, dual, triple, quad, and octo-byte read functions
4. Exponentially growing read-ahead chunk size to reduce amount of buffered data when performing many seeks

There are basically no cases where none of these features are used, so even if we went with the buffered wrapper approach, the benefits would be lost since everything would use the buffered wrapper.

Therefore, to see any advantages, `MediaSource` would need to be extended to support these capabilities. I feel this would be rather complicated for any media source beyond just a big buffer in memory containing the file. So the trade off of using `MediaSourceStream` with an extra copy, I feel, is worth it.

pdeljanov:
[...]

Issue #119, opened 2022-04-03
`MediaSource` should not require `Sync`


Herschel:
#98 added a `Sync` bound to the `MediaSource` trait, but I'm not sure this the best idea. This prevents the use of a `MediaSource` that would otherwise work fine in a single threaded scenario where it is known there will be no data races.

pdeljanov:
Hi @Herschel,

Sorry for the late response.

Yes, in retrospect I do feel like this was not the best change to make. I think I will probably be reverting this for 0.6.

Do you have a `MediaSource` that's running into this problem?

Herschel:
No worries! Generally it's not a huge issue on the user side, and simply involves adding a `Sync` trait bound as appropriate. But because `Sync` is transitive, this can require digging quite deep to ensure the trait is propagated properly. In Ruffle's case, this required us to add some Sync bounds to both our core code and some other decoders (https://github.com/ruffle-rs/ruffle/commit/24f00238a77021bea392e0a91c2125ff6e5d7430). It seems like other users had to do so as well (https://github.com/RustAudio/rodio/pull/412/commits/21adce6194e69218a552ef91c6a5d422c9fab1ed).

So mostly an annoyance in practical cases because most things are `Sync` anyway, but it could be problematic if someone wanted to use the symphonia types in a non-`Sync` scenario, such as in a type with non-thread safe interior mutability like `RefCell`.

PR #236, opened 2023-10-13
Allow non-static sources

### LINKED ISSUE #117: MediaSource without a static life time
So I want to use a media source that depends on an external life time `'a`:
```rs
// My media source implementation:
struct DatasetMediaSource<'a> {
    r: hdf5::ByteReader<'a>, // Implements io::Read and io::Seek, but does not have a static life time
    size: u64,
}
impl<'a> DatasetMediaSource<'a> {
    fn new(ds: &'a hdf5::Dataset) -> Result<DatasetMediaSource<'a>> {
        Ok(DatasetMediaSource {
            r: ds.as_byte_reader()?,
            size: ds.size() as u64,
        })
    }
}
impl<'a> Read for DatasetMediaSource<'a> {
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        self.r.read(buf)
    }
}
impl<'a> Seek for DatasetMediaSource<'a> {
    fn seek(&mut self, pos: SeekFrom) -> std::io::Result<u64> {
        self.r.seek(pos)
    }
}
impl<'a> MediaSource for DatasetMediaSource<'a> {
    fn is_seekable(&self) -> bool {
        true
    }

    fn byte_len(&self) -> Option<u64> {
        Some(self.size)
    }
}
```
However, MediaSourceStream requires a static life time:
```rs
// media_source_stream.rs
pub struct MediaSourceStream {
    /// The source reader.
    inner: Box<dyn MediaSource>, // same as: Box<dyn MediaSource + 'static>
   /// ...
```
Resulting in the error
```rs
let source = Box::new(DatasetMediaSource::new(&ds)?); // type: Box<dyn MediaSource + 'a>
                                              ^^^ borrowed value does not live long enough
```

Does someone have an idea, if I can use such a MediaSource?

a1phyr:
This allows reading data from borrowed buffers and such.

This is a breaking change, because it adds many lifetimes in codebase.

Closes #117

### DIFF
--- a/symphonia-bundle-flac/src/demuxer.rs
+++ b/symphonia-bundle-flac/src/demuxer.rs
@@ -35,8 +35,8 @@ const FLAC_FORMAT_INFO: FormatInfo = FormatInfo {
 };
 
 /// Free Lossless Audio Codec (FLAC) native frame reader.
-pub struct FlacReader {
-    reader: MediaSourceStream,
+pub struct FlacReader<'s> {
+    reader: MediaSourceStream<'s>,
     metadata: MetadataLog,
     tracks: Vec<Track>,
     cues: Vec<Cue>,
@@ -45,9 +45,9 @@ pub struct FlacReader {
     parser: PacketParser,
 }
 
-impl FlacReader {
+impl<'s> FlacReader<'s> {
     /// Reads all the metadata blocks, returning a fully populated `FlacReader`.
-    fn init_with_metadata(source: MediaSourceStream, options: FormatOptions) -> Result<Self> {
+    fn init_with_metadata(source: MediaSourceStream<'s>, options: FormatOptions) -> Result<Self> {
         let mut metadata_builder = MetadataBuilder::new();
 
         let mut reader = source;
@@ -140,18 +140,18 @@ impl FlacReader {
     }
 }
 
-impl Probeable for FlacReader {
+impl Probeable for FlacReader<'_> {
     fn probe_descriptor() -> &'static [ProbeDescriptor] {
-        &[support_format!(FLAC_FORMAT_INFO, &["flac"], &["audio/flac"], &[b"fLaC"])]
+        &[support_format!(FlacReader<'_>, FLAC_FORMAT_INFO, &["flac"], &["audio/flac"], &[b"fLaC"])]
     }
 
-    fn score(_src: ScopedStream<&mut MediaSourceStream>) -> Result<Score> {
+    fn score(_src: ScopedStream<&mut MediaSourceStream<'_>>) -> Result<Score> {
         Ok(Score::Supported(255))
     }
 }
 
-impl FormatReader for FlacReader {
-    fn try_new(mut source: MediaSourceStream, options: FormatOptions) -> Result<Self> {
+impl<'s> FormatReader<'s> for FlacReader<'s> {
+    fn try_new(mut source: MediaSourceStream<'s>, options: FormatOptions) -> Result<Self> {
         // Read the first 4 bytes of the stream. Ideally this will be the FLAC stream marker.
         let marker = source.read_quad_bytes()?;
 
@@ -337,7 +337,7 @@ impl FormatReader for FlacReader {
         Ok(SeekedTo { track_id: 0, actual_ts: packet.ts, required_ts: ts })
     }
 
-    fn into_inner(self: Box<Self>) -> MediaSourceStream {
+    fn into_inner(self: Box<Self>) -> MediaSourceStream<'s> {
         self.reader
     }
 }
--- a/symphonia-bundle-flac/src/demuxer.rs
+++ b/symphonia-bundle-flac/src/demuxer.rs
@@ -150,7 +150,7 @@ impl Probeable for FlacReader<'_> {
     }
 }
 
-impl<'s> FormatReader<'s> for FlacReader<'s> {
+impl<'s> BuildFormatReader<'s> for FlacReader<'s> {
     fn try_new(mut source: MediaSourceStream<'s>, options: FormatOptions) -> Result<Self> {
         // Read the first 4 bytes of the stream. Ideally this will be the FLAC stream marker.
         let marker = source.read_quad_bytes()?;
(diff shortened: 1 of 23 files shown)


pdeljanov:
Thank you, this looks useful! I've tagged this for the next sem-ver breaking release.

a1phyr:
Rebased onto `dev-0.6` branch

a1phyr:
Hi @pdeljanov !

If you eventually want to merge this, could you do it quickly please ? This PR touches a lot of parts of the codebase, so changes on `dev-0.6` create many merge conflicts. 

The clippy failure is unrelated to the PR, it fails on the previous commit too.

pdeljanov:
Hi @a1phyr,

Thanks for diligently updating your PR! I was pushing out a large backlog of changes I had for 0.6, but I'll pause that until we finish with your PR.

It looks like a new clippy warning has appeared after the latest Rust release, but I'll just waive that since it's unrelated to your changes.

I'll try to review this PR sometime in the next few days, but from my initial look-through it seems reasonable enough.

pdeljanov:
Now that I had a chance to sit down with this, I think for the original intent, this is good as-is.

However, I had a thought. Would it be better instead to make the readers generic across `M: MediaSource` instead of just the lifetime `'s`? The original design of Symphonia discounted this idea because I didn't want to pay the monomorphization cost per `MediaSource` type. However, I realize now that this can be avoided by adding an `impl MediaSource for Box<dyn MediaSource>` to the `core::io` module. Then, a Box'd `MediaSource` could be used with dynamic dispatch just as it is right now without forcing it on all users. Does this seem like a good idea?

I understand this is also probably outside the scope you had intended for this PR. I would be happy to merge your PR as-is and I can follow up with another commit, but if you'd like to try and make the change, just let me know.

a1phyr:
If possible, I would prefer making the readers generic across `M: MediaSource` in another PR. I am not sure of the benefits though, I wouldn't do it without measuring the perf improvements and the compile time/binary size cost.

For this PR, I also made 0d177fd51136af104171366d9a29c2c21ad0e732 (but not pushed it to this PR) to avoid the generic lifetime on `FormatReader` by splitting the trait ([full diff](https://github.com/pdeljanov/Symphonia/compare/dev-0.6...a1phyr:Symphonia:lifetimes2)). What do you think ?

pdeljanov:
> I am not sure of the benefits though, I wouldn't do it without measuring the perf improvements and the compile time/binary size cost.

I think the main benefit is that if a user only requires one particular type of `MediaSource` (e.g., `std::fs::File`), then forcing them to box it creates an unnecessary allocation, and also makes all reads on the media source indirect calls.

Of course, the practical overhead of this is basically 0. `MediaSourceStream` was specifically created to amortize the overhead of those read calls by buffering chunks of data at a time. So I don't think we will see much of a performance gain, but I also definitely don't think we will see a loss.

Likewise, assuming the user continues to use a `Box<dyn MediaSource>`, the binary size should not change. My only concern here would be if someone did the following instead of using a box with dynamic dispatch:

```rust
enum SymphoniaFormatReader {
    File(FormatReader<std::fs::File>),
    Buffer(FormatReader<std::io::Cursor>,
}
```
The current design basically prevents them from being able to do this.

We have a number of tickets relating to `MediaSource`s:  #117, #119, #171. I don't think we can implement every suggestion, but I think a generic media source may address most things.

> For this PR, I also made [0d177fd](https://github.com/pdeljanov/Symphonia/commit/0d177fd51136af104171366d9a29c2c21ad0e732) (but not pushed it to this PR) to avoid the generic lifetime on `FormatReader` by splitting the trait ([full diff](https://github.com/pdeljanov/Symphonia/compare/dev-0.6...a1phyr:Symphonia:lifetimes2)). What do you think ?

I'm not sure about this one. Is the motivation primarily for cleanliness?

One issue I see is that that it's not possible to call `into_inner` from a format created by the `Probe` since it returns `Box<dyn FormatReader +'s>` and erases the type of the media source.

> If possible, I would prefer making the readers generic across `M: MediaSource` in another PR. 

How do you feel about just pushing it as a second commit to this PR? If it doesn't work out we can just drop the commit. Usually I just squash all PRs down at merge time so it'll keep the history clean.

a1phyr:
> I'm not sure about this one. Is the motivation primarily for cleanliness?

It is also to avoid having a useless lifetime parameter to `FormatReader`, which is more annoying than useful. Also, removing all genericity from the trait is essential if we want to add a generic type parameter to the readers.


> One issue I see is that that it's not possible to call `into_inner` from a format created by the `Probe` since it returns `Box<dyn FormatReader +'s>` and erases the type of the media source.

I found a way to keep `into_inner` in `FormatReader`. It would be more difficult with generic readers.


>> If possible, I would prefer making the readers generic across `M: MediaSource` in another PR.
>
> How do you feel about just pushing it as a second commit to this PR? If it doesn't work out we can just drop the commit. Usually I just squash all PRs down at merge time so it'll keep the history clean.

I tried doing that but it requires far more work as it needs to touch most functions of the codebase, and would also requires some redesign (many functions actually depend on `MediaSource + ReadBytes + SeekBuffered`). I don't want to do so much work for such a small gain. Sorry :/

pdeljanov:
Hi @a1phyr,

I haven't forgotten about this! Due to real-world reasons, I haven't been able to dedicate any meaningful time to Symphonia.

I still would like to prototype and explore being generic over a `MediaSource`, as well as looking into alternatives to the newly introduced `BuildFormatReader` trait. This will take some time on my end.

I'm still open to merging the PR as it was prior to the `BuildFormatReader` commit since I think it laid a good foundation for my exploration and ultimately resolved the initial problem at hand. Otherwise, I think it may still be a while yet before I can provide a more substantial response.

pdeljanov:
Hi @a1phyr,

After experimenting with making `MediaSource` completely generic and not a boxed trait, the benefits didn't seem to  outweight the additional complexity. So I've merged this PR as-is. Thanks for the work you put into it!

I will look into reworking `Probeable` to take the place of `BuildFormatReader` because `try_new` was only part of `FormatReader` to ensure `support_format!` would always work. I think there is room to streamline things here a bit.

PR #241, opened 2023-11-07
Upgrade bitflags to 2.4.1

GnomedDev:
This is a breaking change, so should be tagged with 0.6.

Replaces the outdated bitflags 1.2.1 with 2.4.1, which reduces downstream duplicate dependencies and allows removing the ChannelsIter as this is automatically added by the bitflags macro.


pdeljanov:
Thanks for doing this. Unfortunately, I already inadvertently did this while working on the `dev-0.6` branch. I appreciate the effort!

GnomedDev:
No problem

PR #244, opened 2023-12-24
No_std support for symphonia-core

atoktoto:
Following the encouragement from https://github.com/pdeljanov/Symphonia/issues/88 and seeing that https://github.com/pdeljanov/Symphonia/pull/91 got stuck I decided to take a shot myself. 

My approach: 
- Get the crate to build with `#![no_std]` and `extern crate alloc;` 
- Make minimal changes in the code base, introduce no extra dependencies 
- Use `core::Error` even though it's still behind `#![feature(error_in_core)]`, switch to Nightly until stabilized ( https://github.com/rust-lang/rust/issues/103765 )
- Start with `core` and `flac` and get it running on actual hardware to see if a simple conversion is viable.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -17,6 +17,7 @@ Thom Chiovoloni <chiovolonit@gmail.com>
 # Please keep this section sorted in ascending order.
 
 aschey [https://github.com/aschey]
+atoktoto [https://github.com/atoktoto]
 BlackHoleFox [https://github.com/blackholefox]
 darksv [https://github.com/darksv]
 dedobbin [https://github.com/dedobbin]
--- a/symphonia-codec-aac/tests/tests.rs
+++ b/symphonia-codec-aac/tests/tests.rs
@@ -31,5 +31,5 @@ fn invalid_channels_aac() {
 
     let err = test_decode(file).unwrap_err();
 
-    assert!(matches!(err, errors::Error::Unsupported(_)));
+    assert!(matches!(err, errors::SymphoniaError::Unsupported(_)));
 }
--- a/symphonia-codec-pcm/src/lib.rs
+++ b/symphonia-codec-pcm/src/lib.rs
@@ -15,9 +15,6 @@
 #![allow(clippy::identity_op)]
 #![allow(clippy::manual_range_contains)]
 
-#[cfg(not(std))]
-extern crate core as std;
-
 extern crate alloc;
 
 use symphonia_core::support_codec;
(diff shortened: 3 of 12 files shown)


pdeljanov:
Hi @atoktoto,

Very interesting work here and looking forward to seeing how far you can get!

I have not done a very thorough read-through of the changes, but I do have some concerns I'd like to bring up before you get in too far. I know this is early days and just proof-of-concept work!

First, I think there needs to be a very firm separation between changes required for `no_std`, and changes required make Symphonia suitable for microcontrollers. These are two different problems, and the latter is going to be much more difficult than the former. It is highly unlikely I will merge a single PR that addresses both at the same time. Hopefully it will become clear why below.

---

In terms of changes to `symphonia-core`:

- I am not comfortable with the copying of `std::io` enums and traits. We have a stricter license than that of the standard library, so this would not be permitted.

- Changing the byte/bit readers from returning `io::Result` to `Result` will likely create a performance hit since the former is smaller than the latter, and these functions are used in extremely hot and inlined code paths. Would need to benchmark this with something like MP3 decoding (uses the bit reader heavily).

---

In terms of the changes you're making to `symphonia-bundle-flac`:

- You cannot disable the CRC checks in `Fragment`. They are used to properly packetize the native FLAC container, and while it may work for some files (particularly error-free ones), that is not a guarantee if the checks are removed.

- `i64` is required to not cause overflows. Intermediate values while decoding may exceed 32 bits. Perhaps with effort it is possible to use a pure 32-bit decode pipeline, but it would need significant validation.

- I'm not sure why removing iterators would be necessary? That is idiomatic Rust and in theory should optimize away.

---

As you can see, changing the decoders to suit small processors can quickly become a major task. For this reason, I think those changes should live in another PR, or a separate set of "micro" versions of the decoders should be created instead. Particularly if you want to use fixed-point instead of floating-point (as the lossy decoders do).

I don't think I am willing to compromise on the functionality, correctness, performance, or maintainability of the current set of decoders to better support microcontrollers. However, making such compromises in a separate complementary set of decoders whether, in-tree or out-of-tree, is okay.

---

Finally, to answer your question:

> Is there a reason why `MediaSource` trait requires `Send + Sync` ? It seems to build without those.

This was a change request made long ago to simplify the effort required to use Symphonia in certain use-cases. It's not a requirement of the library itself, and I will likely be reverting it at the next sem-ver breaking release.

---

I hope this helps guide you!

pdeljanov:
Sounds like you have a pretty solid plan!

It looks like the `Error` part could be the most controversial. To be frank, I have little experience with `no_std` and what the best practices are, so it will probably be a while before I can have any reasonable feedback. Hopefully some other experts can also chime in. :)

For now, I think you could just continue with your plan. As long as you try to keep the changes as uninvasive as possible it should be workable.

PR #251, opened 2024-01-14
Added gated serde support for metadata behind `serde` feature flag

dustins:
I had a use case for needing to (de)serialize the metadata tags. It was a relatively simple code change to add support since there aren't any exotic types being used and any overhead can be hidden behind a feature flag.

I'm relatively new and unfamiliar with much of the project so if there are other obvious parts of core that could benefit please let me know and I could add them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,8 @@ All metadata readers are provided by the `symphonia-metadata` crate.
 
 Serialization/deserialization for metadata can be enbled with the `serde` feature flag.
 
+> **Note:** Serialization/Deserialization is highly dependent on Symphonia versions. There are no guarantees of compatibility between versions.
+
 ## Quality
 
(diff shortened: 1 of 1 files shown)


dustins:
I appreciate your concern but I'm not sure I agree that providing serialization capabilities infers any additional stability. As a library consumer I wouldn't expect any stronger guarantees than what the library API would provide based on versioning scheme.

abique:
At least it should be clarified in the documentation.

dustins:
I'll make that change to reflect the expectations of stability. Ultimately it'll be up to the project to decide with which of our assumptions they align more closely.

dustins:
Just commenting again to see if there is any chance of getting this merged.

dustins:
I'm just circling back on this to see if there is any chance of getting this merged. I was happy to see it at least wasn't closed 😅

PR #252, opened 2024-01-16
Modernize CI

a1phyr:
- Update `actions/checkout`
- Replace `actions-rs/toolchain` with manual install
- Replace `actions-rs/cargo` by manual commands
- Remove redundant "Test documentation" job


pdeljanov:
Hi @a1phyr,

Thanks for this. I'm guessing the motivation was the archival of actions-rs?

Yesterday I cleaned up all the clippy warnings for the 0.5.4 release, would it be possible to have the CI run clippy and show any warnings in GitHub (though, preferably not block the PR)? It's probably the best time to introduce such a step.

qarmin:
Rustup is now builtin into images, so there is no need to use external apps that duplicates rustup functionality

pdeljanov:
Thanks @a1phyr, I think we can run with it. It was only a preference, but it's probably better anyways to have it be mandatory. Let's see how it goes.

I added a commit that updates the contribution guidelines to note that clippy is mandatory now. I also renamed the clippy job to be more clear.

Issue #273, opened 2024-03-24
[mkv] How to obtain `InfoElement::title` (mkv segment title)?


hasezoey:
Currently in symphonia 0.5.4 there does not seem to be a way to obtain mkv's `InfoElement`'s `title` (ie the title that is directly stored in the segment, and not as a metadata element), or is there some way that i didnt find?

ffmpeg for example treats both the same and prints the element's title as a metadata title (unless there is also a metadata title)

some outputs:
```txt
$ cargo run --bin=symphonia-play -- /test/out.mka
|
| // Tracks //
|     [01] Codec:           Vorbis (vorbis)
|          Sample Rate:     48000
|          Duration:        0:01:29.599 (89599)
|          Time Base:       1000000/1000000000
|          Sample Format:   S32
|          Bits per Sample: 32
|          Channel Layout:  Stereo
|          Language:        und
|
| // Tags //
|     [01] ENCODER                      : Lavf60.16.100
|     [02] DURATION                     : 00:01:29.599000000
:

$ ffprobe out.mka                                         
Input #0, matroska,webm, from 'out.mka':
  Metadata:
    title           : Test title
    ENCODER         : Lavf60.16.100
  Duration: 00:01:29.60, start: 0.000000, bitrate: 108 kb/s
  Stream #0:0: Audio: vorbis, 48000 Hz, stereo, fltp
    Metadata:
      DURATION        : 00:01:29.599000000
```

if the `TITLE` metadata is set, symphonia finds it:
```txt
$ cargo run --bin=symphonia-play -- /test/test_title.mka
|
| // Tracks //
|     [01] Codec:           Vorbis (vorbis)
|          Sample Rate:     48000
|          Duration:        0:01:29.599 (89599)
|          Time Base:       1000000/1000000000
|          Sample Format:   S32
|          Bits per Sample: 32
|          Channel Layout:  Stereo
|          Language:        und
|
| // Tags //
|     [01] ENCODER                      : Lavf60.16.100
|     [02] TITLE                        : Test title
|     [03] DURATION                     : 00:01:29.599000000
:

$ ffprobe test_title.mka 
Input #0, matroska,webm, from 'test_title.mka':
  Metadata:
    ENCODER         : Lavf60.16.100
    TITLE           : Test title
  Duration: 00:01:29.60, start: 0.000000, bitrate: 108 kb/s
  Stream #0:0: Audio: vorbis, 48000 Hz, stereo, fltp
    Metadata:
[...]

PR #277, opened 2024-04-25
mkv: get metadata from info element

dedobbin:
As discussed here, https://github.com/pdeljanov/Symphonia/issues/273
I just placed the segment title as a tag. 
When dumping the metadata for an mka file that has both a "tag title" and a "segment title" it looks like this

```
Metadata {
    revisions: [
        MetadataRevision {
            tags: [
                Tag {
                    std_key: None,
                    key: "TITLE",
                    value: String(
                        "real_title",
                    ),
                },
            ],
            visuals: [],
            vendor_data: [],
        },
        MetadataRevision {
            tags: [
                Tag {
                    std_key: None,
                    key: "TITLE",
                    value: String(
                        "tag_title",
                    ),
                },
```

If you think there is a neater way let me know!

### DIFF
--- a/symphonia-format-mkv/src/demuxer.rs
+++ b/symphonia-format-mkv/src/demuxer.rs
@@ -352,7 +352,9 @@ impl FormatReader for MkvReader {
                     segment_tracks = Some(it.read_element_data::<TracksElement>()?);
                 }
                 ElementType::Info => {
-                    info = Some(it.read_element_data::<InfoElement>()?);
+                    let raw_info = it.read_element_data::<InfoElement>()?;
+                    raw_info.copy_metadata_into(&mut metadata);
+                    info = Some(raw_info);
                 }
                 ElementType::Cues => {
                     let cues = it.read_element_data::<CuesElement>()?;
--- a/symphonia-format-mkv/src/segment.rs
+++ b/symphonia-format-mkv/src/segment.rs
@@ -7,7 +7,7 @@
 
 use symphonia_core::errors::{Error, Result};
 use symphonia_core::io::{BufReader, ReadBytes};
-use symphonia_core::meta::{MetadataBuilder, MetadataRevision, Tag, Value};
+use symphonia_core::meta::{MetadataBuilder, MetadataLog, MetadataRevision, Tag, Value};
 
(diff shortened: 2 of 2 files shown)


symphonia-format-mkv/src/segment.rs
```diff
@@ -323,6 +324,19 @@ impl Element for InfoElement {
     }
 }
 
+impl InfoElement {
+    pub fn copy_metadata_into(&self, metadata_log: &mut MetadataLog) {
+        match self.title.clone() {
+            Some(title) => {
+                let mut metadata = MetadataBuilder::new();
+                metadata.add_tag(Tag::new(None, "title", Value::String(title.to_string())));
```
pdeljanov, line 332: I think you should add the `StandardTagKey` for a title here instead of `None`.

symphonia-format-mkv/src/segment.rs
```diff
@@ -323,6 +324,19 @@ impl Element for InfoElement {
     }
 }
 
+impl InfoElement {
+    pub fn copy_metadata_into(&self, metadata_log: &mut MetadataLog) {
+        match self.title.clone() {
```
pdeljanov, line 329: You probably don't need to clone here since you're calling `to_string` later.

symphonia-format-mkv/src/demuxer.rs
```diff
@@ -352,7 +352,9 @@ impl FormatReader for MkvReader {
                     segment_tracks = Some(it.read_element_data::<TracksElement>()?);
                 }
                 ElementType::Info => {
-                    info = Some(it.read_element_data::<InfoElement>()?);
+                    let raw_info = it.read_element_data::<InfoElement>()?;
+                    raw_info.copy_metadata_into(&mut metadata);
+                    info = Some(raw_info);
```
pdeljanov, line 357: Could you do something like this instead (since `info` is never `None`) to make it more concise?

```suggestion
                    info = Some(it.read_element_data::<InfoElement>()?);
                    info.as_ref().unwrap().copy_metadata_into(&mut metadata);
```

pdeljanov (changes requested):

Issue #278, opened 2024-04-26
Mkv/Webm seeking error


GonnaHackIt:
When trying to seek forward in a streamed source only a little bit (like 1-5 seconds, seeking more than about 10 seconds works fine) I get an error that the source is only forward seekable. 

I searched for the cause of it in the [code](https://github.com/pdeljanov/Symphonia/blob/master/symphonia-format-mkv/src/demuxer.rs#L158-L196) and what I think is the problem is that since clusters don't have blocks and we are trying to seek forward in the current cluster, it tries to seek to the start of it, meaning it would go backwards and because the source itself is not seekable it returns the [error](https://github.com/pdeljanov/Symphonia/blob/4f41954de75e2447753f3e522cc7345ec414a703/symphonia-format-mkv/src/ebml.rs#L264).

Prototype solution, put [here](https://github.com/pdeljanov/Symphonia/blob/master/symphonia-format-mkv/src/demuxer.rs#L187):
```rust
let seek_ts = match target_block {
    Some(block) => block.timestamp,
    None => cluster.timestamp
};
let current_ts = self.frames.front().unwrap().timestamp;

if seek_ts > current_ts || ts < current_ts {
    self.iter.seek(pos)?;
}
```
But because I'm not that familiar with the codebase, I don't know how to get the current timestamp if not from the frames which might be empty (eg. the end of track), meaning the unwrap would panic.

If it's not intended behavior and I diagnosed the issue correctly I am willing to try implementing solution for that.

dedobbin:
Hey!

I'm not super familiar with MKV but i studied the code for a bit to get a grasp at what is going on and i understand the problem you are running into, this seems valid yes.

I also think this is the correct place to add the fix. 
I was thinking though, it might be bit more expressive/clear to have a conditional that check if the target cluster is the same as current cluster? The conditional might seem bit obscure if you don't know why it's there.
There doesn't seem a straight forward way to obtain the current cluster though, so a simple comment might be the better approach.

And as for the unwrap, i think if you use a switch statement (or is_some()) to prevent the panic. Something like

```
match self.frames.front(){
  Some(frame) => {
    if seek_ts > frame.current_ts || ts < frame.current_ts {
        self.iter.seek(pos)?;
    }
}
```

I didn't reproduce the problem though, if you have any file/stream that allows me to replicate it and there is no copyright issues with sharing it, would you mind sharing it? Thanks

GonnaHackIt:
Hi @dedobbin!

[...]

PR #284, opened 2024-05-13
mkv: fix seeking error

GonnaHackIt:
This PR fixes the bug discussed [here](https://github.com/pdeljanov/Symphonia/issues/278).


dedobbin:
Since we already went over this quite a bit in the mentioned issue i say this is good to go. If you want to test run it @pdeljanov , the issue also contains some example input and ways to confirm this works in the issue discussion

PR #287, opened 2024-06-10
flac: Allow different sample rate between frames

Gusted:
FLAC encodes for each frame the sample rate of that frame, it is therefore to be expected that the sample rate changes between frames and is valid per FLAC's test vectors. This patch removes the sample rate check in `strict_frame_header_check` and updates the sample rate in the `SignalSpec` of the `AudioBuffer` after decoding a frame.

Ref: https://github.com/ietf-wg-cellar/flac-test-files/blob/main/uncommon/01%20-%20changing%20samplerate.flac

### DIFF
--- a/symphonia-bundle-flac/src/decoder.rs
+++ b/symphonia-bundle-flac/src/decoder.rs
@@ -97,6 +97,11 @@ impl FlacDecoder {
 
         let header = read_frame_header(&mut reader, sync)?;
 
+        // Update the sample rate for this frame.
+        if let Some(sample_rate) = header.sample_rate {
+            self.buf.spec_mut().rate = sample_rate;
+        }
+
         // Use the bits per sample and sample rate as stated in the frame header, falling back to
         // the stream information if provided. If neither are available, return an error.
         let bits_per_sample = if let Some(bps) = header.bits_per_sample {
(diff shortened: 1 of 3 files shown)


Gusted:
There is a test vector that can be used to verify this, but I don't see how the flac crate is tested, so any help would be appreciated.

dedobbin:
Hey ! First of all, thanks a lot for providing the test file, it helps a lot.

I played around with it for a bit, it seems to me the sample rate is not applied correctly and i think this would require a deeper change.

symphonia-play does play the entire file, but the pitch goes up and down, implying the change in samplerate is not 'picked up'.
ffplay does actually play it without pitch changes respecting the changing rate, but when comparing the samples between ffmpeg and symphonia(-check) the samples are different. I find that a bit odd since sample rate should not have effect on how the samples are read, perhaps ffmpeg does a little trick there, but haven't looked deeper into it.

Also side note, vlc just stops playing when the samplerate changes, and (as to be expected) audacity does only apply the initial sample rate.

So yeah to make this work properly it seems to me a deeper change is needed, if this is desirable/viable, i hope @pdeljanov  can say something about that based on this information.

pdeljanov:
Thanks @Gusted for your PR.

Unfortunately, I am not a fan of making the `AudioBuffer` specification mutable since that means channels also becomes  mutable. Changing channels without reallocating the underlying buffer accordingly (which would now be possible) would violate the invariants of `AudioBuffer`.

I think a fix for this may be best targeted for 0.6 (see the `dev-0.6` branch). Not only does 0.6 completely rewrite the audio interfaces, but I think we also need to better specify the `ResetRequired` error since this is a case where the user should be notified that the audio format has changed.

PR #306, opened 2024-07-31
Use a shift instead of div to have bit exact output for MS Adpcm codec.

dtzxporter:
This output now matches FFMPEG/ADPCMEncode.exe's output bit for bit.

### DIFF
--- a/symphonia-codec-adpcm/src/codec_ms.rs
+++ b/symphonia-codec-adpcm/src/codec_ms.rs
@@ -94,11 +94,11 @@ impl AdpcmMsBlockStatus {
     fn expand_nibble(&mut self, byte: u8, nibble: Nibble) -> i32 {
         let nibble = nibble.get_nibble(byte);
         let signed_nibble = signed_nibble(nibble) as i32;
-        let predictor = ((self.sample1 * self.coeff1) + (self.sample2 * self.coeff2)) / 256
-            + signed_nibble * self.delta;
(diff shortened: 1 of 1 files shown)


dedobbin:
Hey,

I played around a bit and wanted to run some files through symphonia-check, i ran into a right shift error though.

```
thread 'main' panicked at symphonia-codec-adpcm/src/codec_ms.rs:97:25:
attempt to shift right with overflow
```

[tone4_adpmc_ms.wav.zip](https://github.com/user-attachments/files/17248249/tone4_adpmc_ms.wav.zip)  
The file i used. It's a tone generated in audacity and converted to adpcm_ms

dtzxporter:
Should be fixed now.

dedobbin:
Ah perfect! running the attached file through symphonia-check gives the same output as with master now.

pdeljanov:
The reason why it overflowed is because evaluating `8 + signed_nibble * self.delta` took [precedence](https://doc.rust-lang.org/reference/expressions.html#expression-precedence) over the shift operation. So the amount shifted was not 8, rather, it was whatever that expression evaluated to, causing an overflow when it was >= 32.

The code below will work without the overhead of `wrapping_shr`.

```rust
let predictor = (((self.sample1 * self.coeff1) + (self.sample2 * self.coeff2)) >> 8) + signed_nibble * self.delta;
```

PR #314, opened 2024-09-30
mp4: Add basic HEVC detection

sscobici:
Very basic video detection of HEVC stream for mp4. Intentionally made this PR small to get feedback.
I'd like to extract raw HEVC NAL units to try to decode them using hardware decoder.


pdeljanov:
Cool!

I actually have a fairly sizeable, yet unpublished, work-in-progress commit for supporting video and subtitle tracks in MP4. It covers some of the things you've added, but fleshed it out more. I also have rudimentary support for detecting AVC via. the AVCDecoderConfigurationRecord, and was going to do the same for the HevcDecoderConfigurationRecord.

If I push these changes, do you want to rebase over it and continue the work? I wasn't planning to go much farther than what I have right now since it's not my main priority. Mainly, I just wanted to see the video and subtitle tracks.

sscobici:
Thanks for your feedback and welcome back, I expected that you also working in similar area. I will wait for your commit. 
I will add missing pieces needed for video bitstream extraction,
I also have some fixes for fragmented mp4 which doesn't have duration in trak.mdia.mdhd.duration.

pdeljanov:
Thanks, I'm looking forward to seeing this developed more! I have pushed my changes in 0c9f1f79e609e59a0847f7ff7d844251e3fd7984.

Even if Symphonia cannot decode video itself, its demuxers should ideally be capable of producing video packets that can be decoded by third-party decoders. I think before that, there will also need to be changes made to `Packet` and possibly other core objects.

PR #322, opened 2024-10-18
core: rename CODEC_ID_NULL and other constants to be unique

sscobici:
Using CODEC_ID_NULL from both audio and video in the same file is not quite convenient and  requires prefixes.
Unique names also remove c-bindings errors


pdeljanov:
Good call

PR #324, opened 2024-10-23
isomp4 take track id from the media source

sscobici:
for isomp4 track id is assigned sequentially from 0, while track id for mkv is taken from the media source.
According to description from the core Track - id should be taken from the media source. This PR fixes track id for mp4.

### DIFF
--- a/symphonia-format-isomp4/src/demuxer.rs
+++ b/symphonia-format-isomp4/src/demuxer.rs
@@ -235,7 +235,11 @@ impl<'s> IsoMp4Reader<'s> {
         }
 
         // Instantiate a `TrackState` for each track in the media.
-        let track_states = (0..moov.traks.len()).map(TrackState::new).collect::<Vec<TrackState>>();
+        let track_states = moov
+            .traks
+            .iter()
+            .map(|trak| TrackState::new(trak.tkhd.id as usize))
+            .collect::<Vec<TrackState>>();
 
         // Instantiate a `Track` for all track states above.
         let tracks = track_states


pdeljanov:
This is going to break in the cases the `id` read from the container does not match identically with the index of the track. The `track_num` in `TrackState` (the field `id` is saved to) is often used to map to an index into a vector to get track-specific data. This vector would need to be changed to a HashMap since the `id` is not guaranteed to be sequential.

sscobici:
> The track_num in TrackState (the field id is saved to) is often used to map to an index into a vector to get track-specific data.
Changing this will be quite a substantial modification with potential performance impacts.

My bad, I haven't observed this. Strange that mkv is not considering this but probably there were no examples with non sequential track ids. I will take further look into this.

ffmpeg AVStream has two fields:
- `index` - zero based index for vector
`id` - the value from the media source

Is it worth introducing second trackid into isomp4 code to not touch existing one, but provide the media source value outside? This will fix mediainfo comparison and also present the values which are read from file.

sscobici:
it seems that track_id(id from file) and track_num(index) are the names that should be used.
Unfortunately these two values are mixed now and this will probably break the parsing for files where track id is not sequential or not starting from 0.

This would be much harder to fix and risky but still possible. Let me know if you want me to do this and what would be the high level approach to allow your easy review?

pdeljanov:
I played around with this, and I think I may take this one since I basically implemented it. It's actually not a big set of changes. Just need to add the `track_id` to `TrackState`, and use it where appropriate (mainly, `seek` and `next_packet`).

pdeljanov:
Fixed. Thanks for noticing the issue though!

PR #326, opened 2024-12-12
Fix clippy lints

a1phyr:


pdeljanov:
Thank you for taking care of these new lints!

PR #327, opened 2024-12-23
riff: fix parsing ffmpeg wave stdout output

sscobici:
Adding back 4 bytes for parent length comparison with u32:MAX as it was reduced by 4 bytes in commit f72ac415f6910e56660247882d78835b81ae6a7e

This fixes` symphonia-check some.mp3` which currently returns error:

> Test interrupted by error: malformed stream: riff: chunk length exceeds parent (list) chunk length

### DIFF
--- a/symphonia-format-riff/src/common.rs
+++ b/symphonia-format-riff/src/common.rs
@@ -92,8 +92,8 @@ impl<T: ParseChunkTag> ChunksReader<T> {
             // input, it may overflow when if added to anything.
             if self.len - self.consumed < len {
                 // When ffmpeg encodes wave to stdout the riff (parent) and data chunk lengths are
-                // (2^32)-1 since the size can't be known ahead of time.
-                if !(self.len == len && len == u32::MAX) {
+                // (2^32)-1 since the size can't be known ahead of time. Parent length was reduced by 4 to read "form type"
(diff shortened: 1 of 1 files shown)


pdeljanov:
Hmm, I see the issue, and this would be a workaround, but I don't think it is a correct fix. Subtracting 4 was just one possible case for AIFF and WAVE, however, it may be different for other RIFF-based formats like AVI which would also be implemented in this crate.

I got hit with a similar issue with MkvReader. I'm rewriting it now so that `EbmlIterator` will take the the length as an option since these magic values always seem to cause bugs in the end.

So, I think the ideal fix here too would be to make the length passed to the chunk reader an `Option<u64>`. In the case where the length is unknown, we'd just pass `None`.

sscobici:
Cancelling this workaround PR. Will wait for your refactoring.

pdeljanov:
I fixed this issue in 5a9bc17af8473dfe2769033b66fa8ebf3b24e590 by passing an `Option<u32>` to the `ChunkReader` and detecting when FFmpeg sets it to `0xffff'ffff`.

Please help test both AIFF and WAVE with this new change.

sscobici:
I don't have too many files but ffmpeg output comparison now works with this commit, while previously it was failing. Thanks

PR #317, opened 2024-10-04
isomp4: fix zero track duration for fragmented mp4

sscobici:
trak.mdia.mdhd.duration sometimes is 0 for fragmented mp4.
There are two files from Dolby Vision for which video packets cannot be extracted because of that:
- iOS_P5_GlassBlowing2_3840x2160@59.94fps_15200kbps.mp4
- P81_GlassBlowing2_3840x2160@59.94fps_15200kbps_fmp4.mp4

Solution: take duration from trak.tkhd.duration
Please review.

Issue #301, opened 2024-07-11
Pack of broken files


qarmin:
Code
```
pub fn parse_audio_file(file_handler: File) -> Result<(), Error> {
    let mss = MediaSourceStream::new(Box::new(file_handler), Default::default());

    let probed = match symphonia::default::get_probe().format(
        &Default::default(),
        mss,
        &Default::default(),
        &Default::default(),
    ) {
        Ok(t) => t,
        Err(_) => {
            return Err(Error::Unsupported(
                "probe info not available/file not recognized",
            ))
        }
    };

    let mut format = probed.format;

    let track = match format
        .tracks()
        .iter()
        .find(|t| t.codec_params.codec != CODEC_TYPE_NULL)
    {
        Some(k) => k,
        None => return Err(Error::Unsupported("not supported audio track")),
    };

    let mut decoder =
        match symphonia::default::get_codecs().make(&track.codec_params, &Default::default()) {
            Ok(k) => k,
            Err(_) => return Err(Error::Unsupported("not supported codec")),
        };

    loop {
        let packet = match format.next_packet() {
            Ok(packet) => packet,
            Err(Error::ResetRequired) => {
                return Err(Error::ResetRequired);
            }
            Err(err) => {
                if let IoError(ref er) = err {
                    // Catch eof, not sure how to do it properly
                    if er.kind() == io::ErrorKind::UnexpectedEof {
                        return Ok(());
                    }
                }
                return Err(err);
            }
        };

        decoder.decode(&packet)?;
    }
}
```
Several crashes, different than previously reported(unwraps, panics etc.)

[Untitled Folder.zip](https://github.com/user-attachments/files/16174190/Untitled.Folder.zip)

sscobici:
Hi @qarmin,

I created a pull request to 0.6 branch which should fix panics from your examples once it will be merged.
What scripts, software you used to generate those test m4a files?

qarmin:
I use cargo-fuzz with corpus and own program(create_broken_files) which randomly modify input files.

Both are used by https://github.com/qarmin/Automated-Fuzzer, which fuzz several programs in CI.

I already changed symphonia commit to https://github.com/pdeljanov/Symphonia/pull/331, which seems to fix most(or even all) isomp4 issues

cargo-fuzz - https://github.com/qarmin/Automated-Fuzzer/actions/runs/12516279820 (results in "minimized..." files)
[...]

PR #332, opened 2024-12-27
isomp4: fix track duration and playback for fragmented files

sscobici:
the approach taken in PR #317 doesn't allow to adjust duration to mdhd timescale, so I reverted the changes and modified only moov atom

Please advise what code I can reuse to avoid/reduce the risk of multiplication overflow during the timescale conversion.
I found some logic in core\units.rs but didn't find it easy to reuse.

### DIFF
--- a/symphonia-format-isomp4/src/atoms/trak.rs
+++ b/symphonia-format-isomp4/src/atoms/trak.rs
@@ -20,8 +20,6 @@ pub struct TrakAtom {
     pub edts: Option<EdtsAtom>,
     /// Media atom.
     pub mdia: MdiaAtom,
-    /// Duration, equal to mdia.mdhd.duration unless it is zero, in which case it is equal to tkhd.duration.
-    pub duration: u64,
 }
 
 impl Atom for TrakAtom {
@@ -45,19 +43,16 @@ impl Atom for TrakAtom {
             }
         }
 
-        let Some(tkhd_atom) = tkhd
+        let Some(tkhd) = tkhd
         else {
             return decode_error("isomp4 (trak): missing tkhd atom");
         };
 
-        let Some(mdia_atom) = mdia
+        let Some(mdia) = mdia
         else {
             return decode_error("isomp4 (trak): missing mdia atom");
         };
 
-        let duration =
-            if mdia_atom.mdhd.duration != 0 { mdia_atom.mdhd.duration } else { tkhd_atom.duration };
-
-        Ok(TrakAtom { tkhd: tkhd_atom, edts, mdia: mdia_atom, duration })
+        Ok(TrakAtom { tkhd, edts, mdia })
     }
 }
--- a/symphonia-format-isomp4/src/demuxer.rs
+++ b/symphonia-format-isomp4/src/demuxer.rs
@@ -17,7 +17,9 @@ use symphonia_core::io::*;
 use symphonia_core::meta::{Metadata, MetadataLog};
 use symphonia_core::units::Time;
 
+use std::collections::HashMap;
 use std::io::{Seek, SeekFrom};
+use std::num::NonZero;
 use std::sync::Arc;
 
 use crate::atoms::{AtomError, AtomIterator, AtomType, ReadAtom};
@@ -46,7 +48,12 @@ pub struct TrackState {
 }
 
 impl TrackState {
-    pub fn make(track_num: usize, trak: &TrakAtom) -> (Self, Track) {
+    pub fn make(
+        track_num: usize,
+        trak: &TrakAtom,
+        media_timespan: &TimeSpan,
+        duration: Duration,
+    ) -> (Self, Track) {
         let mut track = Track::new(trak.tkhd.id);
 
         // Create the codec parameters using the sample description atom.
@@ -55,8 +62,9 @@ impl TrackState {
         }
 
         track
-            .with_time_base(TimeBase::from_recip(trak.mdia.mdhd.timescale))
-            .with_num_frames(trak.duration);
+            .with_time_base(TimeBase::from_recip(media_timespan.timescale))
+            .with_num_frames(media_timespan.duration)
+            .with_duration(duration);
 
         let state = Self {
             track_num,
@@ -96,6 +104,25 @@ struct SampleDataInfo {
     len: u32,
 }
 
+/// A representation of time, defining a duration relative to a specific frequency
+#[derive(Debug)]
+pub struct TimeSpan {
+    pub timescale: NonZero<u32>,
+    pub duration: u64,
+}
+
+impl Default for TimeSpan {
+    fn default() -> Self {
+        Self { timescale: NonZero::new(1).unwrap(), duration: 0 }
+    }
+}
+
+impl TimeSpan {
+    pub fn new(timescale: NonZero<u32>, duration: u64) -> Self {
+        TimeSpan { timescale, duration }
+    }
+}
+
 /// ISO Base Media File Format (MP4, M4A, MOV, etc.) demultiplexer.
 ///
 /// `IsoMp4Reader` implements a demuxer for the ISO Base Media File Format.
@@ -120,7 +147,6 @@ impl<'s> IsoMp4Reader<'s> {
 
         let mut ftyp = None;
         let mut moov = None;
-        let mut sidx = None;
 
         // Get the total length of the stream, if possible.
         let total_len = if is_seekable {
@@ -138,6 +164,8 @@ impl<'s> IsoMp4Reader<'s> {
 
         // Parse all atoms if the stream is seekable, otherwise parse all atoms up-to the mdat atom.
         let mut it = AtomIterator::new(mss, total_len);
+        // Maps each track id to its cumulative duration (TimeSpan) as parsed from the segment index.
+        let mut sidx_timespans: HashMap<u32, TimeSpan> = HashMap::new();
 
         while let Some(header) = it.next_header()? {
             // Top-level atoms.
@@ -149,27 +177,18 @@ impl<'s> IsoMp4Reader<'s> {
                     moov = Some(it.read_atom::<MoovAtom>()?);
                 }
                 AtomType::SegmentIndex => {
-                    // If the stream is not seekable, then it can only be assumed that the first
-                    // segment index atom is indeed the first segment index because the format
-                    // reader cannot practically skip past this point.
-                    if !is_seekable {
-                        sidx = Some(it.read_atom::<SidxAtom>()?);
-                        break;
-                    }
-                    else {
-                        // If the stream is seekable, examine all segment indexes and select the
-                        // index with the earliest presentation timestamp to be the first.
-                        let new_sidx = it.read_atom::<SidxAtom>()?;
+                    let sidx = it.read_atom::<SidxAtom>()?;
 
-                        let is_earlier = match &sidx {
-                            Some(sidx) => new_sidx.earliest_pts < sidx.earliest_pts,
-                            _ => true,
-                        };
-
-                        if is_earlier {
-                            sidx = Some(new_sidx);
-                        }
+                    // calculate total_duration per track from segment indexes
+                    let sidx_timespan = sidx_timespans
+                        .entry(sidx.reference_id)
+                        .or_insert(TimeSpan::new(sidx.timescale, 0));
+                    if sidx_timespan.timescale != sidx.timescale {
+                        return unsupported_error(
+                            "isomp4: different sidx timescale for the same track",
+                        );
                     }
+                    sidx_timespan.duration += sidx.total_duration;
                 }
                 AtomType::MediaData | AtomType::MovieFragment => {
                     // The mdat atom contains the codec bitstream data. For fragmented streams, a
@@ -240,8 +259,7 @@ impl<'s> IsoMp4Reader<'s> {
         let mut moov = moov.unwrap();
 
         if moov.is_fragmented() {
-            // If a Segment Index (sidx) atom was found, add the segments contained within.
-            if sidx.is_some() {
+            if !sidx_timespans.is_empty() {
                 info!("stream is segmented with a segment index.");
             }
             else {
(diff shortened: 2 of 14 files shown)


sscobici:
will mark this as draft, waiting for mp4 refactoring

valarauca:
This doesn't appear to handle edits?

sscobici:
Hi,

Symphonia doesn't currently account for edit lists during MP4 playback. With these edits, a file's duration might be reported as 10s even if it only contains 1s of actual audio/video frames.

I’ve reviewed your proposed changes and noticed there is also a fix in version 0.5. I’ll try adjust this PR to fix MP4 duration in 0.6, though it won't include support for edits or streaming yet. Unless you already have a PR opened.

sscobici:
This PR corrects errors introduced in my previous PR #317.

**Future Work (outside of this PR)**
1. For fragmented file when segment index is not present duration needs to be calculated from Movie Fragments or Track Fragments. This is currently not implemented as it would increase initial parsing time.
2. Currently ignores edits (edts). Demuxer should also be adjusted
3. Does not populate track duration directly; only `num_frames` is populated, maintaining consistency with other demuxers.

**Related Issues**
- Partially fixes #475.

**Credits**
- Applies the fix from 0.5 #440, but implemented at a different level
- Applies the fix from #459 but inside the moov atom

symphonia-core/src/units.rs
```diff
@@ -910,6 +910,24 @@ impl fmt::Display for TimeBase {
...
+        Self { timescale: NonZero::new(1).unwrap(), duration: 0 }
+    }
+}
+
+impl TimeSpan {
+    pub fn new(timescale: NonZero<u32>, duration: u64) -> Self {
+        TimeSpan { timescale, duration }
+    }
+}
+
```
pdeljanov: Let's not add this too core, it's not integrated with anything else in `units`. If there is a need for it to be in core in the future, it can moved later.

symphonia-format-isomp4/src/atoms/moov.rs
```diff
@@ -81,6 +81,12 @@ impl Atom for MoovAtom {
                 }
             }
         }
+        else {
+            // for non fragmented mp4 mdhd.duration must be equal to stts.total_duration
+            for trak in traks.iter_mut() {
+                trak.mdia.mdhd.duration = trak.mdia.minf.stbl.stts.total_duration;
```
pdeljanov: This is for consistency since we don't support edits? If there aren't any edits, then these values should be roughly equal as far as I understand.

symphonia-format-isomp4/src/atoms/trak.rs
```diff
@@ -45,19 +43,16 @@ impl Atom for TrakAtom {
...
         };
 
-        let Some(mdia_atom) = mdia
+        let Some(mdia) = mdia
         else {
             return decode_error("isomp4 (trak): missing mdia atom");
         };
 
-        let duration =
-            if mdia_atom.mdhd.duration != 0 { mdia_atom.mdhd.duration } else { tkhd_atom.duration };
```
pdeljanov, line 59: Ah. This would've been wrong. The TKHD duration is in the movie timescale rather than the track's timescale.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -157,10 +160,21 @@ impl<'s> IsoMp4Reader<'s> {
...
+                            .entry(new_sidx.reference_id)
+                            .or_insert(TimeSpan::new(new_sidx.timescale, 0));
+                        if sidx_timespan.timescale != new_sidx.timescale {
+                            return unsupported_error(
+                                "isomp4: different sidx timescale for the same track",
+                            );
+                        }
+                        sidx_timespan.duration += new_sidx.total_duration;
+
+                        // select the index with the earliest presentation timestamp to be the first.
```
pdeljanov: Do we even need to hold onto `sidx` anymore? It's only used to print an information message if it's `Some`. Instead, you can just see if the hashmap contains atleast 1 entry.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -258,7 +272,13 @@ impl<'s> IsoMp4Reader<'s> {
         let mut track_states = Vec::with_capacity(moov.traks.len());
 
         for (t, trak) in moov.traks.iter().enumerate() {
-            let (track_state, track) = TrackState::make(t, trak);
+            // Determine timescale and duration based on whether the file is fragmented.
+            let timespan = match (moov.is_fragmented(), sidx_timespans.get(&trak.tkhd.id)) {
+                (true, Some(sidx_ts)) => TimeSpan::new(sidx_ts.timescale, sidx_ts.duration),
+                _ => TimeSpan::new(trak.mdia.mdhd.timescale, trak.mdia.mdhd.duration),
+            };
```
pdeljanov: Can we not do a crazy match for this? ;)

How about this:

```rust
let timespan = sidx_timespans
    .get(&trak.tkhd.id)
    .filter(|_| moov.is_fragmented())
    .map(|sidx_ts| TimeSpan::new(sidx_ts.timescale, sidx_ts.duration))
    .unwrap_or_else(|| {
        TimeSpan::new(trak.mdia.mdhd.timescale, trak.mdia.mdhd.duration)
    });
```

symphonia-format-isomp4/src/stream.rs
```diff
@@ -138,21 +138,8 @@ impl StreamSegment for MoofSegment {
...
-            // If a track does NOT end in this segment, then this cannot be the last segment.
-            if seq.first_ts + seq.total_sample_duration < trak.duration {
-                return false;
-            }
-        }
-
-        // All tracks ended.
-        true
+        // cannot determine the total track duration in the fragmented file
+        false
```
pdeljanov, line 142: Are we sure about that? Isn't that why we are summing the duration in the segment index?

pdeljanov (commented):
Thanks for attempting these fixes.

I left some comments for things I'd like to see changed, but also looking for some clarifications to help my understanding.

On an aside, you have 4 PRs open right now, but this is the only non-draft one. Is this the only one you'd like to land in 0.6?

symphonia-core/src/units.rs
sscobici: ok, moved to demuxer

symphonia-format-isomp4/src/atoms/moov.rs
sscobici: Thanks for challenging this. I removed this change. In #440 the code tries to fill duration from stts.total_duration when mdhd.duration is 0, but I'm not sure this is a good idea. We should keep mdhd.duration as it is and just not use during seeking / parsing as it is often different from stts.total_duration (don't know the real reason for this, maybe edits).

mdhd.duration 0 is a valid case for fragmented mp4, because stts.total_duration is also zero in that case.

symphonia-format-isomp4/src/demuxer.rs
sscobici: removed sidx variable, but this PR doesn't store the sidx atoms for further reuse - just calculates the total duration

symphonia-format-isomp4/src/demuxer.rs
sscobici: yes, adjusted.

Seems that format::Track object needs an enhancement.
If we want it to have both num_frames (expressed in media timebase) and duration (expressed in movie timebase), then we need two timebases or use two TimeSpan objects.
MediaTimeSpan can be 2 seconds expressed in media timescale
TrackTimeSpan can be 10 seconds expressed in movie timescale due to edits which can repeat media 5 times.

If we have single timescale we need to transform trak.tkhd.duration into media timescale.

symphonia-format-isomp4/src/stream.rs
sscobici: Some fragmented MP4 files lack a segment index (sidx). In such cases, if the file is seekable, it is possible to parse all movie fragments (moof) to calculate the total track duration (not sure we need this). However, if a fragmented file is non-seekable and lacks a segment index, the duration must be treated as infinite.

        // If all tracks ended in the last segment, then do not try to read anymore segments.
        // Note, there will always be one segment because the moov atom was converted into a segment
        // when the reader was instantiated.
        if self.segs.last().unwrap().all_tracks_ended() {
            return Ok(false);
        }

This check is unreliable for fragmented files. Even after reaching the end of a movie segment, the demuxer must attempt to read the next one—particularly when the total stream duration is unknown

sscobici:
> On an aside, you have 4 PRs open right now, but this is the only non-draft one. Is this the only one you'd like to land in 0.6?

#493 has some minor changes, it will be good if it is merged.
#301 I need to review it, it has some memory and panics fixes, but I need to understand the approach to memory limits in demuxers.
#333 doesn't contain all my local changes but it is very useful for me to test Info against mediainfo and video pts / dts against ffmpeg.

I have some local draft for video PTS calculations but I need more time to get is equivalent to ffmpeg (seems I forgot edits).

----------------
I added setting duration for format::Track in the media timescale. Many files have it populated considering edits.
Seems like symphonia is not good at playing back the files with edits right now and for video files it is often used to sync up audio and video at the start.

symphonia-format-isomp4/src/atoms/trak.rs
sscobici: will try to run some local tests to catch these kind of regressions in future

pdeljanov:
> #493 has some minor changes, it will be good if it is merged.

I'll take a look. Seems pretty minimal and straightforward, so let's see if we can get that one for 0.6.

> #301 I need to review it, it has some memory and panics fixes, but I need to understand the approach to memory limits in demuxers.

This one is based off of the old `dev-0.6` so it's missing the `AtomIterator` rewrite that ended up in `main`. I think a lot of the improvements will have been incorporated in the rewrite. Notably, the `AtomIterator` now acts as a scoped reader, so it's no longer possible to read past the end of an atom. However, there may be so smaller things I missed.

---

I'd like to align on some things regarding timing. In 0.6, we now make a distinction between number of frames and the duration.

The number of frames is the number of decoded audio or video frames, excluding any delay or padding frames (audio only). It is a count, and there is no timebase for it. For users who want to know the total number of decoded frames including discarded, they can add in the delay and padding fields of `CodecParameters`.

The duration is the running time of the track, as stated by the container, in timebase units, excluding any delay or padding frames. Likewise, DTS and PTS are also in timebase units and represent instants along the duration of the track.

Since Symphonia was historically audio-focused, duration and number of frames were often used interchangeably. 
This is because most of the containers did not explicitly state a timebase. So, I chose to use `1 / sample_rate` as the timebase which in turn makes duration and number of frames equal. However, MP4 specifies a timescale, and MKV states durations in nanoseconds. So we can't do this for them.

So, all this to say, for containers that don't specify a timebase, populating the duration of the track with number of frames is completely valid wherever we set the timebase to be `1 / sample_rate`, which is all containers except MKV and MP4.

What may be missing is the concept of an overall media (all track) duration, timebase, etc. I'm not sure how to fit this in with the per-track duration, timebase, etc.

Does this track with your understanding?

sscobici:
> So, all this to say, for containers that don't specify a timebase, populating the duration of the track with number of frames is completely valid wherever we set the timebase to be 1 / sample_rate, which is all containers except MKV and MP4.

I agree. I've added a commit to set the duration as num_frames for other containers; for MKV, there is a better implementation in #494.

> What may be missing is the concept of an overall media (all track) duration, timebase, etc. I'm not sure how to fit this in with the per-track duration, timebase, etc.

I believe general file/stream information should live at the Format level (Readers). We need to define a struct and specify which fields are relevant, and each demuxer will populate them. For now, this PR assumes the track timebase is the media timebase (non general) and handles the conversion for MP4 track durations, which are expressed in the container's general timebase.

sscobici:
Added setting of media_info.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -259,7 +277,45 @@ impl<'s> IsoMp4Reader<'s> {
         let mut track_states = Vec::with_capacity(moov.traks.len());
 
         for (t, trak) in moov.traks.iter().enumerate() {
-            let (track_state, track) = TrackState::make(t, trak);
+            // Determine media timespan and track duration converted to media timescale.
+            let mvhd_timespan = TimeSpan::new(trak.mdia.mdhd.timescale, trak.mdia.mdhd.duration);
```
pdeljanov, line 281: This confused me for a bit because `mvhd_timespan` should be called `mdhd_timespan` or `mdia_timespan` since `mvhd` implies a different timescale all together.

symphonia-format-isomp4/src/stream.rs
pdeljanov: The blocking issue is that I have fragmented files (downloaded from YT) with a segment index, however, with this PR they now return an `Err(EOF)` rather than `Ok(None)` when being piped in via. stdin. Since this is a case with a segment index, it should return `Ok(None)` which means this shouldn't just return `false` in all cases.

What are the cases the existing code became unreliable? Was it the assumption that if a track is not present in the fragment then it must have ended? Are you able to share some samples?

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -259,7 +277,45 @@ impl<'s> IsoMp4Reader<'s> {
...
+                (timespan, duration)
+            }
+            else {
+                // for non fragmented try take duration from tkhd.duration otherwise take duration from media
+                let duration = if trak.tkhd.duration > 0 {
+                    convert_duration(
+                        trak.tkhd.duration,
+                        moov.mvhd.timescale,
+                        mvhd_timespan.timescale,
+                    )
```
pdeljanov, line 309: FYI, this is going to be wrong if there are edits as the `trak.tkhd.duration` field is going to the be duration of all the edits. Otherwise, it should be equal the total duration of all samples. Since we don't support edits, then it would make the most sense to just use the total duration of all samples as that also handles the `trak.tkhd.duration == 0` case. However, you should leave a TODO indicating that here and maybe even the commented out code.

pdeljanov (changes requested):
I think I'd like for this to be the last PR that goes in for 0.6, however, there is a blocking and correctness issue.

I do want to push 0.6 out in a day or two, so let me know if you want me to handle the corrections.

pdeljanov:
Update: I'm quite eager to release 0.6, but without some of the modifications made in this PR MP4 wouldn't be in good shape. So I've been working on a commit based on the changes in this PR (with some modifications based on my comments above) and will likely commit those. I'll make us co-authors on that commit so you will get appropriate attribution.

pdeljanov:
Closing this now since ad5f2d7f117523bdb8e18c3bf7e23fc933afa946 builds upon this PR and with a few adjustments.

sscobici:
thanks for merging.

PR #333, opened 2024-12-30
check: refactored to add check-mode (info, audio). Info mode compares format structure

sscobici:
Refactored symphonia-check to have different run modes: audio, info, ...
It's a draft PR to receive feedback.

**High level overview of changes:**
1. Audio Check:
The audio check functionality was moved with minimal changes. Error handling and messages were adjusted to consistently prefix errors with "FAIL:" when something goes wrong.
2. Info Check:
The info check mode runs mediainfo, producing JSON output (added serde and json dependencies). This JSON output is converted to FormatReader, which is then compared against Symphonia's FormatReader data.
3. FormatReader Comparison:
Implemented a basic comparison for FormatReader data: Generic Format, Track Format, and Track ID/count.
4. Command-Line Changes:
**A command must now be specified** (audio or info), followed by the respective parameters for that command. Updated to the latest version of clap using declarative objects. It is currently not possible to make the audio command the default because clap would interpret parameters as commands, this is a limitation of the library.

```
Usage: symphonia-check.exe <COMMAND>

Commands:
  info   Check Symphonia info output with a reference decoder (mediainfo)
  audio  Check Symphonia audio output with a reference decoder (ffmpeg or other)
  help   Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

**Observations:**
1. VideoCodecId implements `From<FourCc>`, which might falsely imply that you can create a VideoCodecId from a FourCc and retrieve it later. Implementing a generic passthrough for unknown codecs in symphonia-check not easy because the existing VideoCodecId values (from well_known) do not correspond to valid FourCc values. There is no straightforward way to pass unsupported codec IDs from the demuxer through FormatReader. Adding a placeholder string to pass through what was found, alongside CODEC_ID_NULL_VIDEO, would be beneficial.
2. FORMAT_INFO constants are private and not exposed, requiring external applications to define short and long names for well-known CODEC_IDs and FORMAT_IDs. Besides exposing numeric constants for well-known formats and codecs, it would be useful to expose their short and long names to avoid redundant definitions in applications.
3. Recent changes to chapter and tag parsing result in file processing failures:
```
FAIL: malformed stream: mkv: missing chapter display language
or 
apetag invalid, file doesn't play, unexpected end of file error
```
Tags and menu chapters provide additional information and should not block playback when invalid.

**Todo: (to be addressed in separate PRs)**
1. add video check mode that uses ffprobe packet output to validate PTS / DTS, packet length, packet position, extra_info. Perform checks without decoding video data.
2. Add `pos` and `size` fields to the Packet struct to support better video checks.
3. Add more info mode validations: duration, size, bitrate, language, dolby vision, etc
4. Add detection of some old video formats "VC-1".


symphonia-check/src/main.rs
```diff
@@ -12,450 +12,146 @@
 // in the remaining fields with default values.
 #![allow(clippy::needless_update)]
 
-use std::fs::File;
-use std::path::Path;
-use std::process::{Command, Stdio};
+use std::{io::ErrorKind, process::Command};
```
sscobici, line 15: will split into two lines import in next PR

sscobici:
Added more parameters for comparison, some info issues spotted in different containers. Duration is the most common one, where the difference is in ms, will take a look later to decide if duration comparison should be without milliseconds.

sscobici:
closing, will keep it for my internal testing.

PR #336, opened 2025-01-03
Hoist lints from individual packages to workspace

caass:
By hoisting the lint settings out from individual packages to the root `Cargo.toml` (and using `lints.workspace = true` in the individual crates), lints can be applied uniformly across all crates.


pdeljanov:
Thanks, for this!

I'm glad this option exists to make things more consistent across all of our crates. It looks like `symphonia-core` wasn't using `#![warn(rust_2018_idioms)]` like all the other crates and it's causing the CI errors. Do you want to fix those?

pdeljanov:
Closing due to inactivity.

PR #340, opened 2025-01-16
flac, isomp4, mkv: optimize length detection by using byte_len() instead of seeking

roderickvd:
Replace seeking to end of file with `byte_len()` call to get total stream length. This improves performance when streaming files by avoiding an unnecessary seek operation. The change maintains the same functionality while being more efficient, especially for remote/streaming files where seeks are expensive.

As used in [`pleezer`](https://github.com/roderickvd/pleezer), previously seeking would block the application until the full file was downloaded, _even if the position seemed to was already buffered_. With this change, such seeks are nearly instantaneous.

Optionally, we could first try `byte_len()` and fall back to seeking to the end when `byte_len()` is not available.

The actual change is just three lines, but my editor automatically fixed some minor code formatting issues with else clauses. Nothing more than `rustfmt`.

### DIFF
--- a/symphonia-bundle-flac/src/demuxer.rs
+++ b/symphonia-bundle-flac/src/demuxer.rs
@@ -10,7 +10,9 @@ use std::io::{Seek, SeekFrom};
 use symphonia_core::support_format;
 
 use symphonia_core::codecs::{CodecParameters, VerificationCheck, CODEC_TYPE_FLAC};
-use symphonia_core::errors::{decode_error, seek_error, unsupported_error, Result, SeekErrorKind};
+use symphonia_core::errors::{
+    decode_error, seek_error, unsupported_error, Error, Result, SeekErrorKind,
+};
 use symphonia_core::formats::prelude::*;
 use symphonia_core::formats::util::{SeekIndex, SeekSearchResult};
 use symphonia_core::io::*;
@@ -69,8 +71,7 @@ impl FlacReader {
                         let mut new_index = SeekIndex::new();
                         read_seek_table_block(&mut block_stream, header.block_len, &mut new_index)?;
                         index = Some(new_index);
-                    }
-                    else {
+                    } else {
                         return decode_error("flac: found more than one seek table block");
                     }
                 }
@@ -204,8 +205,7 @@ impl FormatReader for FlacReader {
                 // known, the seek cannot be completed.
                 if let Some(sample_rate) = params.sample_rate {
                     TimeBase::new(1, sample_rate).calc_timestamp(time)
-                }
-                else {
+                } else {
                     return seek_error(SeekErrorKind::Unseekable);
                 }
             }
@@ -229,7 +229,8 @@ impl FormatReader for FlacReader {
             // lower bound is set to the byte offset of the first frame, while the upper bound is
             // set to the length of the stream.
             let mut start_byte_offset = self.first_frame_offset;
-            let mut end_byte_offset = self.reader.seek(SeekFrom::End(0))?;
+            let mut end_byte_offset =
+                self.reader.byte_len().ok_or(Error::SeekError(SeekErrorKind::Unseekable))?;
 
             // If there is an index, use it to refine the binary search range.
             if let Some(ref index) = self.index {
@@ -265,13 +266,11 @@ impl FormatReader for FlacReader {
 
                 if ts < sync.ts {
                     end_byte_offset = mid_byte_offset;
-                }
-                else if ts >= sync.ts && ts < sync.ts + sync.dur {
+                } else if ts >= sync.ts && ts < sync.ts + sync.dur {
                     debug!("seeked to ts={} (delta={})", sync.ts, sync.ts as i64 - ts as i64);
 
                     return Ok(SeekedTo { track_id: 0, actual_ts: sync.ts, required_ts: ts });
-                }
-                else {
+                } else {
                     start_byte_offset = mid_byte_offset;
                 }
             }
@@ -366,8 +365,7 @@ fn read_stream_info_block<B: ReadBytes + FiniteStream>(
 
         // Add the track.
         tracks.push(Track::new(0, codec_params));
-    }
-    else {
+    } else {
         return decode_error("flac: found more than one stream info block");
     }
 
--- a/symphonia-format-isomp4/src/demuxer.rs
+++ b/symphonia-format-isomp4/src/demuxer.rs
@@ -8,7 +8,9 @@
 use symphonia_core::{errors::end_of_stream_error, support_format};
 
 use symphonia_core::codecs::CodecParameters;
-use symphonia_core::errors::{decode_error, seek_error, unsupported_error, Result, SeekErrorKind};
+use symphonia_core::errors::{
+    decode_error, seek_error, unsupported_error, Error, Result, SeekErrorKind,
+};
 use symphonia_core::formats::prelude::*;
 use symphonia_core::io::{MediaSource, MediaSourceStream, ReadBytes, SeekBuffered};
 use symphonia_core::meta::{Metadata, MetadataLog};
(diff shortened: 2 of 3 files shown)


pdeljanov:
Hi @roderickvd, yeah, that's the way it should've been. Nice find!

The reason you got so many formatting changes is because we need the nightly rustfmt to use the brace style option. Are you able to run that with `cargo +nightly fmt`?

roderickvd:
> Hi @roderickvd, yeah, that's the way it should've been. Nice find!

Cool.

Some questions after thinking over seeking some more:

- Currently, when `byte_len == None` then seeking will bail out with a `Unseekable` error for all these formats. Looking at the MP3 implementation, that's true for coarse seeking, but not for accurate seeking. With accurate seeking in MP3 an attempt is made to rewind to the beginning and then do a linear search. Would this be a viable strategy for the other formats also?

- For MP3, it's a hard-fail when `SeekMode == Coarse` and `byte_len == None`. I guess it could be changed to soft-fail by falling back to accurate (linear) seeking. Less obvious to the user, but perhaps more in line with the user intent to "please just seek"?

- The other formats only implement accurate seeking, using a binary search. Would MP3's coarse preseeking be viable to implement for them too? I'm no audio format expert, just from a cursory code read imagine it would be possible.

- Beyond ergonomics, note that Rodio wraps all sources in a [`ReadSeekSource`](https://github.com/RustAudio/rodio/blob/master/src/decoder/read_seek_source.rs) struct with `byte_len` always set to `None`. So, in all transparency, this PR will break Rodio's seeking for everything but MP3. This is a shortcoming in Rodio that I intend to contribute to. In parallel I'm wondering if a "rewind-and-linear-search" fallback strategy could prevent such foot-shooting.

Can move to a discussion to not lead this PR astray too much.

> The reason you got so many formatting changes is because we need the nightly rustfmt to use the brace style option. Are you able to run that with `cargo +nightly fmt`?

Sure, just did so in f143245. Let me know if you want me to squash.

> By the way, please keep an eye on the `dev-0.6` branch as that will be the future of Symphonia in a little while.

Yes, will do. You think this could make its way into a `0.5.5` release? For `pleezer` and `librespot` it would be great if I could keep depending on a "stable" release while `dev-0.6` work is ongoing.

roderickvd:
> * Beyond ergonomics, note that Rodio wraps all sources in a [`ReadSeekSource`](https://github.com/RustAudio/rodio/blob/master/src/decoder/read_seek_source.rs) struct with `byte_len` always set to `None`. So, in all transparency, this PR will break Rodio's seeking for everything but MP3. This is a shortcoming in Rodio that I intend to contribute to. In parallel I'm wondering if a "rewind-and-linear-search" fallback strategy could prevent such foot-shooting.

Refactoring Rodio's decoder to take a `byte_len` here: https://github.com/RustAudio/rodio/pull/697

pdeljanov:
> Some questions after thinking over seeking some more:
> 
> * Currently, when `byte_len == None` then seeking will bail out with a `Unseekable` error for all these formats. Looking at the MP3 implementation, that's true for coarse seeking, but not for accurate seeking. With accurate seeking in MP3 an attempt is made to rewind to the beginning and then do a linear search. Would this be a viable strategy for the other formats also?

There's no timing information in a MP3, so to seek in a sample accurate manner the only way is to count samples from the beginning of the file. In contrast, coarse seeking estimates the average bitrate of the file and then uses it to estimate where the reader should seek to. If the underlying source is unseekable, then we assume the `seek` function would just not work. So we can only seek forward from the current position by again consuming MPEG frames and counting.

Other formats are generally better in this regard. They either contain a table mapping timestamp-to-position, or each packet is timestamped so we can binary search. If I recall correctly, most of these readers will perform a seek in two stages: seek to a rough position guided by the table or binary search algorithm, then refine the position by reading sequentially until the desired timestamp is found. Usually, if the source is unseekable, it just falls back to the second stage so forward seeks work atleast. If this fallback is missing anywhere, then yes, we could do it!

> * For MP3, it's a hard-fail when `SeekMode == Coarse` and `byte_len == None`. I guess it could be changed to soft-fail by falling back to accurate (linear) seeking. Less obvious to the user, but perhaps more in line with the user intent to "please just seek"?

I believe this should work, but it may be failing if `byte_len == None` but `is_seekable == true`. There is an unstated assumption that `byte_len != None` if `is_seekable == true` because you could always seek to get the byte length. The documentation for implementers of the `MediaSource` trait could be improved here.

> * The other formats only implement accurate seeking, using a binary search. Would MP3's coarse preseeking be viable to implement for them too? I'm no audio format expert, just from a cursory code read imagine it would be possible.

Hmm, possibly. Though I guess this would be an optimistic optimization: no need to seek multiple times for a binary search if a first-try guess gets close enough.

> * Beyond ergonomics, note that Rodio wraps all sources in a [`ReadSeekSource`](https://github.com/RustAudio/rodio/blob/master/src/decoder/read_seek_source.rs) struct with `byte_len` always set to `None`. So, in all transparency, this PR will break Rodio's seeking for everything but MP3. This is a shortcoming in Rodio that I intend to contribute to. In parallel I'm wondering if a "rewind-and-linear-search" fallback strategy could prevent such foot-shooting.

Noted!

Since I imagine most people use Symphonia through Rodio, this would be a pretty bad breakage. It would be great if Rodio could push an update with your fixes before this lands in a Symphonia release.

> > By the way, please keep an eye on the `dev-0.6` branch as that will be the future of Symphonia in a little while.
> 
> Yes, will do. You think this could make its way into a `0.5.5` release? For `pleezer` and `librespot` it would be great if I could keep depending on a "stable" release while `dev-0.6` work is ongoing.

Yeah, we could include this in 0.5.5. I was planning to cut a release before `master` becomes 0.6.

yara-blue:
On the Rodio side we have run into users who found setting a `bytes_len` and `seekable` separately confusing. I tend to agree. I have tried to improve our facade around `symphonia` however I run into issues.

Specifically: 
- only some *codecs* require `bytes_len` for seeking.
- `bytes_len = None` and `seekable =  true` leads to panics for some codecs.
- we only know the *codec* once we have a track.
- we can only make a track after creating a `MediaSourceStream`.
- when creating `MediaSourceStream` we have to know whether to set `seekable`

I can work around this by sharing an atomic between our implementation of `MediaSourceStream` and the code that uses it to figure out the *codec*. That feels very wrong to me.

An alternative would be modifying symphonia such that `seekable` no longer exists instead a seek on an unseekable stream returns an error.

roderickvd:
> * only some _codecs_ require `bytes_len` for seeking.

This is a bug that this PR intends to address. Like @pdeljanov stated:

> There is an unstated assumption that byte_len != None if is_seekable == true because you could always seek to get the byte length. The documentation for implementers of the MediaSource trait could be improved here.

So we should assume _all_ codecs require `byte_len` for seeking. Which will be the case when this PR gets merged.

yara-blue:
> This is a bug that this PR intends to address. Like @pdeljanov stated:

ahh thanks for the clarification, I missed that. Given that is the case what would be the objection to enabling seek-able always and only if bytes_len.is_some()?

edit:
I have a test in rodio, which with this PR, shows only m4a, flac and ogg need `bytes_len.is_some()`. Does that mean this PR is incomplete? That is what originally confused me.

roderickvd:
> ahh thanks for the clarification, I missed that. Given that is the case what would be the objection to enabling seek-able always and only if bytes_len.is_some()?

`is_seekable` implies support for **random-access** seeking. Sources that could have their byte length known, yet only be forward-seekable include pipes, buffers from `stdin` or a read-once network stream.

Sources that could have its byte length known, yet not support random-access seeking could be HTTP downloads without 

> edit: I have a test in rodio, which with this PR, shows only m4a, flac and ogg need `bytes_len.is_some()`. Does that mean this PR is incomplete? That is what originally confused me.

That could be. The scope of this PR is to no longer use `reader.seek(SeekFrom::End(0))` to get the byte length. In the discussion above you see that indeed not all codecs have seek implemented equally. But that's beyond the scope of this PR.

roderickvd:
*bump* what's the hold-up to getting this merged?
I can retarget dev-0.6 if you want.

pdeljanov:
@roderickvd , nothing was. I must've missed it when catching up on outstanding PRs. Thanks! I will cut a 0.5.5 release this weekend.

roderickvd:
Awesome thanks.

PR #342, opened 2025-01-18
riff: fix clippy warnings, fix overflow with u16 for frames_per_block calculation

sscobici:
should fix builds for other PRs.


sscobici:
added another fix for riff, in addition to clippy warnings

pdeljanov:
Thanks!

Issue #196, opened 2023-03-25
bug: ADTS timebase missing


erikas-taroza:
Hello.

I use this crate as a dependency in my project. A user ran into an issue with a missing timebase for an ADTS file.

[Here is the sample file I was provided.](https://github.com/erikas-taroza/simple_audio/files/11067975/silent.zip)

Using `symphonia-play`, the file is able to be played. However, there is no information about the position or duration (as expected with a missing timebase). Playing the file in VLC works as expected (duration + position are shown).

pdeljanov:
I think there are probably two problems:

1. The time base is missing. This should be pretty simple to fix since for audio it's just `1 / sample_rate` and the ADTS header provides it.

2. A missing time base just means the timestamps can't be converted to a time in seconds. However, you would still see the position and duration represented as sample count counts (played and total). The second issue is that ADTS, not unlike MP3, doesn't actually record the total duration. Instead, it must be estimated by finding the average bitrate and dividing the file size by the bitrate. This is a bit more work, but the MP3 demuxer is a good starting point. MP3 also sometimes has XING, LAME, or VBRI frames embedded into the stream to more accurately provide the duration. If there are similar frames for ADTS I'd be interested in hearing about them if someone is aware of any.  

I believe @dedobbin is already investigating this issue, but I figured I should make a note of these observations here.

erikas-taroza:
> 1. The time base is missing. This should be pretty simple to fix since for audio it's just `1 / sample_rate` and the ADTS header provides it.

Yes, this is the workaround that my project is currently using.



> I believe @dedobbin is already investigating this issue, but I figured I should make a note of these observations here.

Sounds good :)

dedobbin:
Hey! Been some time, sorry about that. 
In `symphonia-codec-aac/src/adts.rs` i added `.with_time_base(TimeBase::new(1, header.sample_rate))` but this was not enough.
`.with_n_frames(n_frames);` is also necessary, the trouble is here that each frame has it's own header with a variable size, so it's a bit tricky the find the amount.  

Lazy solution would be to just go through the entire file to find the individual sizes and determine the amount using that, but that's not great.

I found another decent solution by just parsing a few blocks and using the average size to estimate the total length. It's obviously not super accurate but it doesn't seem to be an uncommon approach. 
I need to call MediaSource::byte_len for it though, which is costly. 

Do you agree with this approach @pdeljanov  ?

BSteffaniak:
[...]

PR #343, opened 2025-01-18
aac: refactor approximate frame count, fuzzy fixes

sscobici:
refactored aac aproximate_frame_count() method. It had some issues which causes panic. Should fix cases from https://github.com/qarmin/Automated-Fuzzer/actions/runs/12516290494 (REPORTS___SYMPHONIA*)

1. renamed total_len to remaining_len because this is not the length of the whole file, but it was considered such in some places.
2. first sample point to start from the beginning, file can contain a single frame
3. check step to be bigger than 0 to avoid panic
4. skip the rest of the frame after the header is read to avoid looking for sync word in the audio data
5. break if reading 100 frames overlap next sample point

I haven't tested on VBR aac to observe any improvements, It was https://github.com/pdeljanov/Symphonia/issues/196#issuecomment-1960507107that approximate method is not perfect there

### DIFF
--- a/symphonia-codec-aac/src/adts.rs
+++ b/symphonia-codec-aac/src/adts.rs
@@ -394,7 +394,7 @@ impl FormatReader for AdtsReader<'_> {
 
 fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Option<u64>> {
     let original_pos = source.pos();
-    let total_len = match source.byte_len() {
+    let remaining_len = match source.byte_len() {
         Some(len) => len - original_pos,
         _ => return Ok(None),
     };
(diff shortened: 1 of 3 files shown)


symphonia-codec-aac/src/adts.rs
```diff
@@ -428,34 +428,48 @@ fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Opt
...
+                    parsed_n_frames += 1;
+                    n_bytes += u64::from(header.frame_len);
 
-                parsed_n_frames += 1;
-                n_bytes += u64::from(header.frame_len);
+                    // skip frame to avoid meeting sync word in the audio data and for quick sync()
+                    cur_pos += u64::from(header.frame_len);
+                    if source.seek(SeekFrom::Start(cur_pos)).is_err() {
+                        break;
+                    }
```
pdeljanov, line 456: Nice catch for an optimization! However, I think `ignore_bytes` will work even better since the MSS will decide if it should just consume from it's internal ring buffer or resort to an actual seek. In other words, you can just `ignore_bytes(header.frame_len)`.

symphonia-codec-aac/src/adts.rs
```diff
@@ -428,34 +428,48 @@ fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Opt
     else {
         // The number of points to sample within the stream.
         const NUM_SAMPLE_POINTS: u64 = 4;
+        const NUM_FRAMES: u32 = 100;
```
pdeljanov, line 431: A slightly more descriptive name.

```suggestion
        const NUM_FRAMES_PER_SAMPLE: u32 = 100;
```

pdeljanov (changes requested):

sscobici:
Thanks for suggestion.
ignore_bytes should ignore only the frame.payload because sync word and header were already read.
added another exit point in a loop to break if searching within the sample point overflow the next sample point. Can happen when no sync word found in a very broken file.

symphonia-codec-aac/src/adts.rs
```diff
@@ -428,34 +428,55 @@ fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Opt
...
+                    if source.ignore_bytes(next_frame_pos - source.pos()).is_err() {
+                        break;
+                    }
+                }
+
+                // Reading frames from the next sample point could lead to reprocessing already calculated frames.
+                // Also, if no sync word was found, better to avoid redundant searches.
+                if next_frame_pos > new_pos + step {
+                    break;
+                }
```
pdeljanov, line 469: Suggestion: Remove this since it wouldn't be required anymore.

symphonia-codec-aac/src/adts.rs
```diff
@@ -428,34 +428,55 @@ fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Opt
...
 
-            for _ in 0..=100 {
-                let header = match AdtsHeader::read(&mut source) {
-                    Ok(header) => header,
-                    _ => break,
-                };
+                    // if reading NUM_FRAMES_PER_SAMPLE_POINT frames overflow the next sample point position then break
+                    if next_frame_pos > new_pos + step {
+                        break;
+                    }
```
pdeljanov, line 454: Suggestion: Remove this since it wouldn't be required anymore.

symphonia-codec-aac/src/adts.rs
```diff
@@ -428,34 +428,55 @@ fn approximate_frame_count(mut source: &mut MediaSourceStream<'_>) -> Result<Opt
...
-        // Skip the first sample point (start of file) since it is an outlier.
-        for new_pos in (original_pos..total_len - step).step_by(step as usize).skip(1) {
-            let res = source.seek(SeekFrom::Start(new_pos));
-            if res.is_err() {
-                break;
-            }
+        // file can be small enough and not have enough NUM_FRAMES_PER_SAMPLE_POINT, but we can still read at least one frame
+        if step > 0 {
+            for new_pos in (original_pos..(original_pos + remaining_len)).step_by(step as usize) {
+                if source.seek(SeekFrom::Start(new_pos)).is_err() {
```
pdeljanov, line 438: Suggestion: Add a test here to make sure we never go backwards or continue from the current position.

```rust
// Skip the current sample point if a previous sample read past it.
if new_pos <= source.pos() {
    continue;
}
```

pdeljanov:
> Thanks for suggestion. ignore_bytes should ignore only the frame.payload because sync word and header were already read. added another exit point in a loop to break if searching within the sample point overflow the next sample point. Can happen when no sync word found in a very broken file.

I added some additional suggestions that could help clean-up the implementation after these changes. Let me know what you think.

pdeljanov:
Thanked, merged!

PR #344, opened 2025-01-18
core: advance reader's position if pattern cannot be found, fixes out of memory from #301

sscobici:
This addresses out-of-memory errors mentioned in the following comment: https://github.com/pdeljanov/Symphonia/issues/301#issuecomment-2569727809.

Issue description:

For a big buffer where the pattern was not found at the end, the reader advanced its position to the end of the buffer.
For a buffer smaller than the pattern length, the reader did not advance its position and entered into infinite loop while reading tags.

This commit ensures that the reader's position is always advanced, resolving the issue.

### DIFF
--- a/symphonia-core/src/io/buf_reader.rs
+++ b/symphonia-core/src/io/buf_reader.rs
@@ -57,21 +57,22 @@ impl<'a> BufReader<'a> {
         // return the remainder of the buffer or scan_length bytes, which ever is shorter, we return
         // that here.
         if remaining < pattern.len() || scan_len < pattern.len() {
+            self.pos = end;
             return Ok(&self.buf[start..end]);
         }
 
-        let mut j = start;
-        let mut i = start + pattern.len();
+        let mut i = start;
+        let mut j = start + pattern.len();
 
-        while i < end {
-            if &self.buf[j..i] == pattern {
+        while j < end {
+            if &self.buf[i..j] == pattern {
                 break;
             }
-            i += align;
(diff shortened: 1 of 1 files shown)


symphonia-core/src/io/buf_reader.rs
```diff
@@ -57,21 +57,22 @@ impl<'a> BufReader<'a> {
         // return the remainder of the buffer or scan_length bytes, which ever is shorter, we return
         // that here.
         if remaining < pattern.len() || scan_len < pattern.len() {
+            self.pos = end;
             return Ok(&self.buf[start..end]);
         }
 
-        let mut j = start;
-        let mut i = start + pattern.len();
+        let mut i = start;
```
pdeljanov, line 64: Why were these renamed? Usually the outer loop variable is `i`, and `j` is the inner loop. It adds noise the commit. Or am I missing something subtle?

sscobici: sorry it was hard to read for me that i is greater than j.
I agree with your statement about outer and inner loop, but there is a single loop here and buf[j..i] looked too wierd for me.

np I can remove this, just confirm you want to remove noise.

pdeljanov: Okay, if it's too weird to you then it's probably too weird to others as well. I merged it as is.

It was looking strange to me because there is a single loop, but `j` was being used in the loop condition instead of `i`.

PR #345, opened 2025-01-18
adpcm: report unsupported error for more than 2 channels

sscobici:
fixes panic

### DIFF
--- a/symphonia-codec-adpcm/src/lib.rs
+++ b/symphonia-codec-adpcm/src/lib.rs
@@ -88,6 +88,10 @@ impl AdpcmDecoder {
         };
 
         let spec = if let Some(channels) = &params.channels {
+            if channels.count() > 2 {
+                return unsupported_error("adpcm: up to two channels are supported");
+            }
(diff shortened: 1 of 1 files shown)


symphonia-codec-adpcm/src/lib.rs
```diff
@@ -140,7 +140,7 @@ impl AdpcmDecoder {
                     decode_stereo(&mut stream, buffers, frames_per_block)?;
                 }
             }
-            _ => unreachable!(),
+            _ => return unsupported_error("adpcm: up to two channels are supported"),
```
pdeljanov, line 143: This isn't the correct place to return an unsupported error, especially since the channel count is constant after instantiation.

Please move this validation check to `try_new` before we create the audio buffer with > 2 channels (line 91 seems good).

pdeljanov (changes requested):

sscobici:
fixed.

pdeljanov:
Thanks again!

PR #346, opened 2025-01-18
metadata: validate min chunk length for jpeg

sscobici:
fixes panic


symphonia-metadata/src/utils/images.rs
```diff
@@ -82,6 +82,9 @@ fn parse_jpeg(mut reader: BufReader<'_>) -> Result<ImageInfo> {
 
             return Ok(info);
         }
+        if chunk_len < 2 {
+            return decode_error("meta (jpeg): invalid chunk length");
+        }
```
pdeljanov, line 87: Looking further into JPEG, it looks like there are some chunks that don't actually have a length (they're just markers essentially) [^1]. These are `chunk_type`s `0xd0` to `0xd9`, inclusive.

After line 66, can you add a `continue` if any of these chunk types are encountered?
Then, please move this check to after line 67 where `chunk_len` is read, since a length < 2 is just invalid entirely.

[^1]: https://github.com/corkami/formats/blob/master/image/jpeg.md

pdeljanov (changes requested):

sscobici:
fixed

pdeljanov (approved):

pdeljanov:
Thank you!

PR #347, opened 2025-01-18
core: add pts, dts

sscobici:
adding pts (former ts) and dts.

This was done in two commits: the first for renaming `ts` to `pts`, and the second for adding `dts` (i64), which can be negative for video.
Audio crates will not need to pass `dts` when creating the Packet using the`new` methods. Video packets are planned to be created with `new_from_boxed_slice_v`, which will include a `dts` parameter. `new_from_boxed_slice_v` is not present in this commit, but I have changes for MP4 to include DTS. Waiting for MP4 refactoring if you plan to do it.

For some reason, only the `ogg` crate uses `packet.ts` directly, which is why it was impacted as well.
Can the `ogg` crate be rewritten to use internal structures instead?


pdeljanov:
I'm quite bothered by `pts` and `dts` being differing types (`u64` vs. `i64`). It feels like `pts` should also just be `i64` so that there is no `pts` that can't also be a `dts`. I imagine this will break a bunch of stuff...

sscobici:
>  I imagine this will break a bunch of stuff...

Not a lot if we agree with approach to hide the complexity in Packet constructor methods, so that
`new_from_boxed_slice` - for audio, - I can rename it to `new_from_boxed_slice_a` if you want.
`new_from_boxed_slice_v`  - for video

I added a commit for pts: i64
I will squash to a single commit - after review.

pdeljanov:
Hey, sorry for the long absence.

Regarding this change, I think it makes sense for pts and dts to be `i64`. However, it feels wrong to me to do the silently lossy unsigned-to-signed conversion.

I think the preferred solution here would be to update the demuxers to use `i64` instead of `u64`. This could be risky for regressions, but it's probably worth it and may simplify some things.

sscobici:
Hi, I'd like to clarify the scope of this PR.
Initial intention was to just add DTS and rename TS to PTS and keep it as u64.

Adding proper support for negative PTS is a much bigger change, which might also impact how the seek works and duration is calculated. Probably it make sense only for video demuxer I don't know, but video with negative PTS should be a very rare thing, may be it will be useful for streaming in future, I don't know yet.

I propose to discuss the design of negative PTS support separately as I'm not sure this is a priority. Moving silent conversion from common to demuxers (u64 to i64) still keep it a silent conversion, but introducing u64.try_into().unwrap() has a performance impact.

PS. I have also other pending changes, which I'd like to commit. which allowed me to playback the video using symphonia demuxer in MPC-BE windows video player (https://github.com/Aleksoid1978/MPC-BE) through FFI 

I observed that current public API structures are not FFI friendly and have performance impact, but that is another discussion about the design of the library.

pdeljanov:
Hi @sscobici, you're correct that it is a bigger change. It doesn't need to be addressed in this PR alone.

On the audio side, a negative PTS makes sense to represent encoder delay. Right now we have "trim" values in the `Packet` to indicate to the decoder what to remove, but with a signed PTS it could theoretically just trim anything before PTS = 0 instead. Much more elegant and simplifies some things.

I'm okay with your workaround of adding a helper function to `Packet` to silently do the conversion for now if we mark those helpers as deprecated from the start. However, before 0.6 is released, I feel like the switch-over to signed timestamps should be done. Then, we don't need expose those helpers in any publicly released version, nor do we need to create churn by removing them in a 0.7.

pdeljanov:
Closing this as 666b2132a716dde43ebcc1e8ee897d7a3eddd03c now supports signed PTS and DTS.

PR #348, opened 2025-01-18
mkv: detect parent overrun after the header parsing

sscobici:
Declaring a constant for EbmlElementHeader<EbmlSchema> doesn't allow to use EbmlElementHeader::MIN_SIZE without specifying the scheme which I considered too complex for a constant. I can define it globally in the file or inside the Schema if you want.  Currently it is a 2 with a comment

I added skipping of a byte to put reader position to the end of parent. Does it make sense, taking into consideration that it will be skipped later anyway?

For element header seeking I propose to rewind the stream to the `parent_end` vs `header.pos`, which should be similar to as everything was ok. Let me know your preference.

### DIFF
--- a/symphonia-format-mkv/src/ebml.rs
+++ b/symphonia-format-mkv/src/ebml.rs
@@ -150,6 +150,9 @@ impl<S: EbmlSchema> Clone for EbmlElementHeader<S> {
 }
 
 impl<S: EbmlSchema> EbmlElementHeader<S> {
+    /// Minimum EBML header size.
+    pub const MIN_SIZE: u64 = 2;
+
     /// Read an EBML element header from the stream.
     pub(crate) fn read<R: ReadBytes>(reader: &mut R, depth: u8, schema: &S) -> Result<Self> {
         // Read the variable width element ID.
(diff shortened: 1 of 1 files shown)


pdeljanov:
> Declaring a constant for EbmlElementHeader doesn't allow to use EbmlElementHeader::MIN_SIZE without specifying the scheme which I considered too complex for a constant. I can define it globally in the file or inside the Schema if you want. Currently it is a 2 with a comment

This should be pretty simple since everything is already parameterized by a schema, `S`. The min. size is really a property of EBML, and not the schema, however, we don't have any lower abstraction for an EBML header other than `EbmlElementHeader<S>`.

Did you try this?

```rust
impl<S: EbmlSchema> EbmlElementHeader<S> {
    /// Minimum EBML header size.
    pub const MIN_SIZE: u64 = 2;
    // ...
```

```rust
if parent_end - pos < EbmlElementHeader::<S>::MIN_SIZE {
    // ...
```


> I added skipping of a byte to put reader position to the end of parent. Does it make sense, taking into consideration that it will be skipped later anyway?

The skip later will yield a debug message which is useful for indicating a strangely muxed file. I wouldn't call `ignore_bytes` so we can see this message.

> For element header seeking I propose to rewind the stream to the `parent_end` vs `header.pos`, which should be similar to as everything was ok. Let me know your preference.

This is okay.

pdeljanov:
Thanks again!

PR #351, opened 2025-02-05
fix(isomp4): Update MIME type for m4a files to audio/mp4

roderickvd:
Changed the MIME type from "audio/m4a" to "audio/mp4" in the format descriptor, as audio/mp4 is the standard MIME type for MPEG-4 audio files according to IANA specifications.


pdeljanov:
Thanks!

PR #358, opened 2025-02-22
no_std for symphonia-core: switch to core and alloc

atoktoto:
Hello again,

This is the first, least intrusive step in the direction of supporting no_std build environment. I switched all `std::` imports to their `core::` or `alloc::` counterparts leaving the rest untouched. This change should be 100% backwards compatible and make the further changes less noisy. 

I intend to follow the same path as described in https://github.com/pdeljanov/Symphonia/pull/244 but in smaller steps this time around :)


sscobici:
use dev-0.6 branch instead of master as it has many changes.
I'm still not sure if no_std is in plan for 0.6.

PR #360, opened 2025-02-22
no_std symphonia-core: replace io::Result with crate::Result

atoktoto:
This is my attempt at reducing dependency on `std::io` in `symphonia-core` with the aim of unblocking future `no_std` compatibility. In this pull request I replaced all `io::Result` with `crate::Result`.

I intend to follow the plan laid out in the discussion here: https://github.com/pdeljanov/Symphonia/pull/244

### DIFF
--- a/symphonia-codec-aac/src/aac/codebooks.rs
+++ b/symphonia-codec-aac/src/aac/codebooks.rs
@@ -6,6 +6,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use symphonia_core::io::{vlc::*, ReadBitsLtr};
+use symphonia_core::Result;
 
 use lazy_static::lazy_static;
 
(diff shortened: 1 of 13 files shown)


baylesj:
The big this is that this patch needs to be re-targeted onto `main` now that `0.6` has been released.

The second thing is that while I think for `no_std` we definitely need to remove our dependency on `std::io::Result`, now that `Error` and `Result` have moved to core (e.g. see https://doc.rust-lang.org/core/error/index.html) instead of std, I think there's some refactoring to do in `symphonia-core`. I think that effort to update [symphonia-core/src/errors.rs](https://github.com/pdeljanov/Symphonia/blob/main/symphonia-core/src/errors.rs) is likely separate from landing this CL, which I think should now be possible since 0.6 is stable.

PR #361, opened 2025-03-06
feat: derive Eq and PartialEq for SeekMode

roderickvd:
This is useful to be able to do:

```rust
if seek_mode == SeekMode::Accurate { ... }
```

instead of the less readable:

```rust
if matches!(seek_mode, SeekMode::Accurate) { ... }
```

Ran into this as we're improving sample-accurate seeking in Rodio: https://github.com/RustAudio/rodio/pull/697


roderickvd:
Unrelated CI failure would be fixed by #378.

pdeljanov:
Sure, makes sense to me. Thanks!

PR #364, opened 2025-03-19
Fix Default implementation for TimeBase

1c3t3a:
A TimeBase represents symphonia's conversion factor between time. It does not make sense for it to have a factor of 0, which is also checked by its constructor TimeBase::new. Unfortunately the Default implementation for TimeBase is derived, yielding a value of 0 for both u32 members (numerator and denominator), which is probably not what we want. This change switches the default to be 1.

Also fixes a typo in the doc comment for TimeStamp.

### DIFF
--- a/symphonia-core/src/units.rs
+++ b/symphonia-core/src/units.rs
@@ -9,7 +9,7 @@
 
 use std::fmt;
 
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks!

Issue #325, opened 2024-11-09
Cannot decode AAC stream from internet radio


dzubybb:
Hi,

I'm trying to play internet radio stream from [https://n-11-21.dcs.redcdn.pl/sc/o2/radio357/live/radio357_pr.livx?preroll=0](https://n-11-21.dcs.redcdn.pl/sc/o2/radio357/live/radio357_pr.livx?preroll=0).

Content type of stream is `audio/aac`. Debug output from `cvlc` (managed to play the stream but the sound was choppy) suggests it's mpeg4audio with AAC in ADTS format :
```
es demux debug: detected format mp4a
mpeg4audio demux packetizer debug: running MPEG4 audio packetizer
main demux packetizer debug: using packetizer module "mpeg4audio"
mpeg4audio demux packetizer debug: detected ADTS format
mpeg4audio demux packetizer: AAC channels: 2 samplerate: 48000
main decoder debug: using audio decoder module "faad"
main input debug: `https://n-11-21.dcs.redcdn.pl/sc/o2/radio357/live/radio357_pr.livx?preroll=0' successfully opened
main input debug: Stream buffering done (1002 ms in 2 ms)
pulse audio output debug: changed buffer metrics: maxlength=4194304, tlength=46080, prebuf=0, minreq=15360
pulse audio output debug: connected to sink RDPSink
main audio output debug: output 'f32l' 48000 Hz Stereo frame=1 samples/8 bytes
main volume debug: looking for audio volume module matching "any": 2 candidates
main volume debug: using audio volume module "float_mixer"
main audio output debug: input 'f32l' 48000 Hz Stereo frame=1 samples/8 bytes
main audio filter debug: looking for audio filter module matching "scaletempo": 16 candidates
scaletempo audio filter debug: format: 48000 rate, 2 nch, 4 bps, fl32
scaletempo audio filter debug: params: 30 stride, 0.200 overlap, 14 search
scaletempo audio filter debug: 1.000 scale, 1440.000 stride_in, 1440 stride_out, 1152 standing, 288 overlap, 672 search, 2400 queue, fl32 mode
main audio filter debug: using audio filter module "scaletempo"
main audio output debug: conversion: 'f32l'->'f32l' 48000 Hz->48000 Hz Stereo->Stereo
main audio output debug: conversion pipeline complete
main audio output debug: conversion: 'f32l'->'f32l' 48000 Hz->48000 Hz Stereo->Stereo
main audio output debug: conversion pipeline complete
[...]

PR #367, opened 2025-03-25
Exiting if invalid streaming content

skoky:
I had to fork Symphonia and do a small change in the mpeg decoder. I am using this with https://crates.io/crates/stream-download crate. This is using Rodio and Symphonia to decode radio stream. But I am getting the `invalid mpeg audio header` warning from time to time. The streamed radio is not playing and I can't detect this erorr to restart the stream. I need o have the error propagated back to Rodio and streaming-donload stream to have it restarted. Wiht this fix it works fine in my fork. Can you include similar fir in next version please? 

The fix is related to the https://github.com/pdeljanov/Symphonia/issues/325 If you made some autodotection improvements in 0.6-dev branch its even better, but returning the errro is really necessary to handle the streaming restart. Thanks

### DIFF
--- a/symphonia-bundle-mp3/src/demuxer.rs
+++ b/symphonia-bundle-mp3/src/demuxer.rs
@@ -20,7 +20,7 @@ use crate::header::{self, MAX_MPEG_FRAME_SIZE, MPEG_HEADER_LEN};
 
 use std::io::{Seek, SeekFrom};
 
-use log::{debug, info, warn};
+use log::{debug, error, info, warn};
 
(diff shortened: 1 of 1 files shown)


pdeljanov:
Sorry, this is not an appropriate fix. The reason why this case does not return an error is because if the stream is broken/corrupt Symphonia should re-synchronize to the next valid MPEG frame and then continue.

First, please test the radio stream with the `symphonia-play` test application: `curl -s https://radio.station.com/stream | symphonia-play -`.

I'm going to close this PR, but if you find that the issue reproduces with the test application, please feel free to open an issue.

PR #376, opened 2025-04-25
AAC: Fix panic in `approximate_frame_count`

Mrmaxmeier:
For short files where step == 0 the `.step_by(step as usize)` call panics. In these situations an approximate frame count is probably not meaningful, so I've added it as an extra edge case.


Mrmaxmeier:
Ah, looks like this was already addressed in the dev branch in 851a80abd6237ecbd12ebe7b147c6ce4e493ffa1

PR #385, opened 2025-05-28
Caf fixes

acieslewicz:
These changes are needed for caf files containing ima4 audio. Main change is moving away from compresses/uncompressed to variable\fixed packet parsing. I've tested with the sample files that worked previously as well as the caf ima4 file I have locally and it seems to work. Let me know if I missed anything or should change something!

### DIFF
--- a/symphonia-format-caf/src/chunks.rs
+++ b/symphonia-format-caf/src/chunks.rs
@@ -220,8 +220,8 @@ impl AudioDescription {
         Ok(result)
     }
 
-    pub fn format_is_compressed(&self) -> bool {
-        self.bits_per_channel == 0
+    pub fn is_variable_packet_format(&self) -> bool {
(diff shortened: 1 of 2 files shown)


pdeljanov:
I managed to do some limited testing of this. Though, I have very few CAF files. I muxed ALAC audio into CAF and that seemed to work after I fixed a bug in the ALAC decoder. I also managed to test some CAF with raw PCM audio.

Unfortunately the CI timed out on this, but there is one clippy issue. Do you mind fixing that?

acieslewicz:
Thanks for testing it! I also ran into the issue of having a limited collection of CAF files and I'm not sure if I can even share those. Clippy issue should be fixed.

pdeljanov:
Thanks @acieslewicz, merged! You could email me some sample files if you'd like.

Issue #330, opened 2024-12-27
Version 0.6 Development Status


pdeljanov:
Hey all,

As some may have noticed, most of the development activity on Symphonia is occurring in the `dev-0.6` branch. This is because I believe the 0.5.x series has reached a developmental dead-end where it is not possible to satisfactorily fix many of Symphonia's issues due to API constraints. Therefore, I've decided to focus my limited time on completing version 0.6 instead.

Version 0.6 includes many breaking changes to support popular requests such as making it easier to get audio out of Symphonia, or returning `None` when the end of the media is reached, but it also includes forward looking enhancements such as stubbing out video and subtitle decoding support.

### Development Plan

My development plan is as follows:

1. For the time being, I will be keeping 0.6 development in the `dev-0.6` branch until there are no more breaking changes being made.
2. Once the APIs are stabilised, I will cut a final 0.5.5 release from the mainline branch that includes some fixes for annoyances people have had. 
3. Once 0.5.5 is released, I will merge the 0.6 development branch into mainline and begin more thorough testing.
4. Once 0.6 has been tested (including some fuzzing time), it will be released.

### Timeline

I am hopeful for an early 2025 release, but some help from the community would improve the odds!

### Help Wanted

You can help the development and release of version 0.6 by picking up any of these tasks:

- [ ] Add fuzz testing harnesses to each format and codec crate.
- [ ] Begin playing around with version 0.6 and submitting feedback and suggestions on the new APIs. **Note: At this time, some of the new APIs are still in flux, but you can still get an idea, and different perspectives will be helpful to me.**
- [ ] Read, critique, fix, or update documentation.

I will add to this list over time...

Thanks all!

crutchcorn:
Hey! I'm using the 0.6 branch against my [very very very small Tauri application](https://github.com/crutchcorn/moonsound), since it fixes many issues with FLAC currently

Just wanted to say that you're crushing it and I'm extremely happy as a downstream consumer of this package and that the API changes to 0.6 were pretty easy to adapt to

I'm still extremely green to Rust in general, so might not be a huge help, but if I end up with some spare cycles and find something critically blocking I'm happy to jump in

caass:
[...]

PR #392, opened 2025-06-30
Upgrade bitflags from 1.2.1 to 2.9.1

baylesj:
This patch upgrades the bitflags dependency to the latest v2 version. This upgrade includes some breaking changes, so the audio.rs file has been edited to work with the new version of the bitflags crate.

This makes managing dependencies easier downstream (I work primarily in the Chromium codebase) but also has benefits purely for Symphonia, since bitflags v2 focuses on improved safety and genericity.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -18,6 +18,7 @@ Thom Chiovoloni <chiovolonit@gmail.com>
 # Please keep this section sorted in ascending order.
 
 aschey [https://github.com/aschey]
+baylesj [https://github.com/baylesj]
 BlackHoleFox [https://github.com/blackholefox]
 darksv [https://github.com/darksv]
 dedobbin [https://github.com/dedobbin]
--- a/symphonia-core/Cargo.toml
+++ b/symphonia-core/Cargo.toml
@@ -29,7 +29,7 @@ opt-simd = [
 
 [dependencies]
 arrayvec = "0.7.1"
-bitflags = "1.2.1"
+bitflags = "2.9.1"
 bytemuck = "1.7"
 lazy_static = "1.4.0"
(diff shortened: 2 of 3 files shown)


baylesj:
I noticed that `master` appears to be for maintaining `0.5.x` versions, and active development is on `dev-0.6`. I think this could be good to land for both  -- should we land in `dev-0.6` first and then cherry pick to `master`?

baylesj:
Looks like the dev branch is already on bitflags v2 -- so maybe this is not needed?

Is there any information available about the release schedule for 0.6?

baylesj:
This is a breaking change but relatively minor. Would be nice to include in the final `0.5.5` version discussed in https://github.com/pdeljanov/Symphonia/issues/330, I'll also push a patch to update dependencies on dev-0.6.

baylesj:
@pdeljanov mind taking a look?

pdeljanov:
Hmm, I think the issue with this is that the MSRV for 2.9.1 (and any 2.x by the look of it), is greater-than the MSRV of Symphonia. So, technically, we could break peoples' builds if they depended on the lower MSRV. Of course, this doesn't matter for the 0.6 branch. Is this a major headache for you?

baylesj:
http://crbug.com/428224817 is the motivating bug.

Our goal is to generally only include one version of each package in Chrome, and currently Chrome pulls in bitflags version 1.3.2 because of Symphonia and 2.9.1 for other packages.

So, I wouldn't say a major headache. I would like to roll patches and updates upstream when it seems appropriate, and I also understand that Symphonia has other embedders than Chromium so not everything we want makes sense for the library.

Certainly something that we can just plan for the 0.6 branch.

PR #393, opened 2025-06-30
Update dependencies

baylesj:
This patch updates dependent packags to latest. No breaking changes.


pdeljanov:
What's the motivation to do this? This is making the version requirements for dependencies _more_ strict for end-users. It's probably only worthwhile to update `symphonia-play` and `symphonia-check`'s unique dependencies like `cpal` and `clap`.

baylesj:
That makes sense. I don't think this patch is needed as is, and I'll close this request.

PR #394, opened 2025-07-01
Cleanup Cargo.tomls and licenses (master branch)

### LINKED ISSUE #391: Include LICENSE files in the published symphonia-* crates
While the Cargo.toml specifies a license (or licenses), crates also typically include a copy of that license (or licenses) as LICENSE files. This github repo contains the license files but the actual symphonia crates do not.

Can the crates please include the license files moving forward? I'm happy to help however I can as well.

As a consumer of third party open source libraries, Chromium is aiming to ensure we include the license files in our own source tree for all of our third-party dependencies, and the symphonia crates are some of a few for which they are missing. Chromium issue: [crbug.com/369075726](https://crbug.com/369075726)

baylesj:
This patch makes it so that the MPl-2.0 license is bundled with each crate (as is best practice for license files). However, cargo does NOT currently have a better solution than copying the license to each crate's folder (see discussion here:
https://github.com/rust-lang/cargo/issues/13328), so this patch copies the LICENSE file to each crate's folder.

This patch also consolidates a lot of the package metadata into the root Cargo.toml.

resolves #391

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,16 @@ members = [
     "symphonia-utils-xiph",
     "symphonia-check",
 ]
+
+# Common workspace attributes shared by all crates.
+[workspace.package]
+authors = ["Philip Deljanov <philip.deljanov@gmail.com>"]
+version = "0.5.4"
+license = "MPL-2.0"
+license-file = "LICENSE"
+homepage = "https://github.com/pdeljanov/Symphonia"
+repository = "https://github.com/pdeljanov/Symphonia"
+categories = ["multimedia", "multimedia::audio", "multimedia::encoding"]
+keywords = ["audio", "codec", "decoder", "multimedia", "media"]
+edition = "2018"
+rust-version = "1.53"
--- a/symphonia-bundle-flac/Cargo.toml
+++ b/symphonia-bundle-flac/Cargo.toml
@@ -1,16 +1,16 @@
 [package]
 name = "symphonia-bundle-flac"
-version = "0.5.4"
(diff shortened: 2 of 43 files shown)


Cargo.toml
```diff
@@ -22,3 +22,16 @@ members = [
...
+authors = ["Philip Deljanov <philip.deljanov@gmail.com>"]
+version = "0.5.4"
+license = "MPL-2.0"
+license-file = "LICENSE"
+homepage = "https://github.com/pdeljanov/Symphonia"
+repository = "https://github.com/pdeljanov/Symphonia"
+categories = ["multimedia", "multimedia::audio", "multimedia::encoding"]
+keywords = ["audio", "codec", "decoder", "multimedia", "media"]
+edition = "2018"
+rust-version = "1.53"
```
hasezoey, line 37: FYI to my knowledge the `workspace.package` syntax required Rust version 1.64 or higher according to [The Cargo Book: Workspaces](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table)

pdeljanov (changes requested):
As mentioned by @hasezoey in the comments, the workspace syntax is only supported on Rust 1.64+ and the MSRV for the 0.5.x series is 1.53. Since we consider increasing the MSRV a semver-breaking change, we cannot use that syntax in `master` at this time. Therefore, I think this PR can only add the `LICENSE` files to each crate.

pdeljanov:
Closing since #391 is fixed.

baylesj:
Thank you for selecting the relevant changes and getting them landed.

PR #395, opened 2025-07-01
Cleanup Cargo.tomls and licenses

baylesj:
This patch makes it so that the MPl-2.0 license is bundled with
    each crate (as is best practice for license files). However, cargo
    does NOT currently have a better solution than copying the license
    to each crate's folder (see discussion here:
    https://github.com/rust-lang/cargo/issues/13328), so this patch
    copies the LICENSE file to each crate's folder.
    
    This patch also consolidates a lot of the package metadata into the
    root Cargo.toml.


symphonia-core/Cargo.toml
```diff
@@ -28,12 +30,11 @@ opt-simd = [
 ]
 
 [dependencies]
-bitflags = "2.4.2"
+arrayvec = "0.7.1"
+bitflags = "1.2.1"
```
pdeljanov, line 34: What?

pdeljanov:
Closing this as this as it has been resolved.

PR #400, opened 2025-08-08
Fix Fuzzing crash by using Result() instead of assert!

baylesj:
This patch fixes a downstream fuzzing crash in Chrome, caused by asserting on input data.

Wherever possible, returning an Error Result is greatly preferred, since crashing on user data may cause unexpected application crashes or potentially other issues.

See related Chromium issue: https://g-issues.chromium.org/issues/433557303

### DIFF
--- a/symphonia-bundle-flac/src/decoder.rs
+++ b/symphonia-bundle-flac/src/decoder.rs
@@ -180,7 +180,7 @@ impl FlacDecoder {
                             bits_per_sample,
                             self.buf
                                 .plane_mut(i)
-                                .ok_or(Error::DecodeError("invalid plane data"))?,
+                                .ok_or(Error::DecodeError("flac: unexpected channel assignment"))?,
                         )?;
                     }
                 }
@@ -190,7 +190,7 @@ impl FlacDecoder {
                     let (left, side) = self
                         .buf
                         .plane_pair_mut(0, 1)
-                        .ok_or(Error::DecodeError("invalid plane data"))?;
+                        .ok_or(Error::DecodeError("flac: unexpected channel assignment"))?;
 
                     read_subframe(&mut bs, bits_per_sample, left)?;
                     read_subframe(&mut bs, bits_per_sample + 1, side)?;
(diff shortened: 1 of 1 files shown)


baylesj:
I think this fix is also part of a larger scoped discussion around how we want to handle issues with input data. In Chromium, we want the decoder to error whenever there is an issue with user data, instead of crash, which has a negative impact on the user. In this library, it would be ideal if, generally speaking, issues with user data ended up in an error Result, and asserts were limited to developer/logic errors.

baylesj:
@pdeljanov mind taking a look?

symphonia-bundle-flac/src/decoder.rs
```diff
@@ -175,29 +175,44 @@ impl FlacDecoder {
             match header.channel_assignment {
                 ChannelAssignment::Independant(channels) => {
                     for i in 0..channels as usize {
-                        read_subframe(&mut bs, bits_per_sample, self.buf.plane_mut(i).unwrap())?;
+                        read_subframe(
+                            &mut bs,
+                            bits_per_sample,
+                            self.buf
+                                .plane_mut(i)
+                                .ok_or(Error::DecodeError("invalid plane data"))?,
```
pdeljanov, line 183: For error strings, we have a convention of prefixing it with the component/crate. Let's change all these error strings to `flac: unexpected channel assignment`.

pdeljanov (changes requested):
Nice find, let's just adjust the error string!

baylesj:
Applied feedback. Thank you for the review!

pdeljanov:
> I think this fix is also part of a larger scoped discussion around how we want to handle issues with input data. In Chromium, we want the decoder to error whenever there is an issue with user data, instead of crash, which has a negative impact on the user. In this library, it would be ideal if, generally speaking, issues with user data ended up in an error Result, and asserts were limited to developer/logic errors.

I agree.

We should always return a decode error. Unwrap should only be used after we pre-validate that it, in-fact, can be unwrapped. In the newer areas of the codebase unwraps usually have a "Safety:" comment/justification associated with them to help auditing, but that's a more recent addition.

PR #401, opened 2025-08-30
mkv: include channels in codec params

aschey:
The mkv demuxer already supports parsing the channel information, but it wasn't being added to the codec params. According to the [spec](https://www.matroska.org/technical/elements.html), there was a `ChannelPositions` element used to denote the channel positions, but it's [listed as reclaimed](https://datatracker.ietf.org/doc/draft-ietf-cellar-matroska/21/) (not in use), so I'm not sure if there's a defined way to map the specific channel positions. 

Please let me know if I missed anything. Thanks!

### DIFF
--- a/symphonia-format-mkv/src/codecs.rs
+++ b/symphonia-format-mkv/src/codecs.rs
@@ -16,6 +16,7 @@ use symphonia_common::mpeg::video::{
 };
 use symphonia_common::xiph::audio::flac::{MetadataBlockHeader, MetadataBlockType};
 use symphonia_core::audio::sample::SampleFormat;
+use symphonia_core::audio::Channels;
 use symphonia_core::codecs::audio::well_known::{CODEC_ID_FLAC, CODEC_ID_VORBIS};
 use symphonia_core::codecs::audio::AudioCodecParameters;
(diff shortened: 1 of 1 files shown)


symphonia-format-mkv/src/codecs.rs
```diff
@@ -66,6 +67,11 @@ fn make_audio_codec_params(
     }
 
     codec_params.with_sample_rate(audio.sampling_frequency.round() as u32);
+    codec_params.with_channels(match audio.channels.into() {
+        1 => layouts::CHANNEL_LAYOUT_MONO,
+        2 => layouts::CHANNEL_LAYOUT_STEREO,
+        n => Channels::Discrete(n as u16),
+    });
```
pdeljanov, line 74: This is probably not universally true.

The channels element only specifies the number of channels in the track, but doesn't dictate their positions. It's probably most correct to make all cases discrete.

It's always been the case that to get the most accurate codec parameters, the user needs to open the decoder and call `codec_info` on it since only a decoder can accurately determine all codec parameters.

pdeljanov (changes requested):

pdeljanov:
Thanks!

PR #407, opened 2025-10-09
fix: Calculate end delay for non-seekable ogg streams

### LINKED ISSUE #404: Chained ogg streams sometimes have dicontinuities after performing a decoder reset
I'm trying to play some internet radio streams and I've noticed that sometimes there's a slight glitch or pop when the next track starts and the ogg demuxer returns `ResetRequired`. I've [forked symphonia-play](https://github.com/aschey/Symphonia/blob/stream-test/symphonia-play/src/main.rs) (branched off dev-0.6) to be able to play HTTP streams and I can reproduce the issue with this. 

It doesn't always happen - sometimes it's barely noticeable, but sometimes there's a pretty audible pop. This is the stream that I've been testing with: `cargo run --bin=symphonia-play http://play.global.audio/radio1rock_low.ogg`.

aschey:
Resolves #404.

Currently, the end delay for ogg streams is only calculated for seekable streams. This can cause discontinuities for non-seekable streams like internet radio that are designed to play gaplessly. Calling `inspect_end_page` for every page that's read ensures the end delay is still calculated for non-seekable streams. If the end delay was calculated previously, this is a no-op because of the check for `self.end_bound.is_some()` at the beginning of `inspect_end_page`.

`InspectState` was also moved to a struct field since the value needs to persist through multiple calls to `read_page`.

### DIFF
--- a/symphonia-format-ogg/src/logical.rs
+++ b/symphonia-format-ogg/src/logical.rs
@@ -48,6 +48,7 @@ pub struct LogicalStream {
     start_bound: Option<Bound>,
     end_bound: Option<Bound>,
     gapless: bool,
+    state: InspectState,
 }
 
 impl LogicalStream {
@@ -63,6 +64,7 @@ impl LogicalStream {
             start_bound: None,
             end_bound: None,
             gapless,
+            state: Default::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


aschey:
Closing since this was addressed separately.

PR #411, opened 2025-10-17
core: Replace std with core and alloc, where possible

keks:
This tries to pick up where #358 left off, in an effort towards #88. This PR only does obvious things that are easy to review. I will try to look at the others later.

The remaining uses of std are:
- `std::collections::HashMap`. Unfortunately this is not yet in alloc, and I am not sure when it will be. See https://github.com/rust-lang/rust/issues/27242
- `std::io` and `std::fs` for obvious reasons.
- `std::error`. While `error_in_core` was stabilized in 1.81.0, it looks like the Rust version given in the Cargo.toml is older than that, so I left it for now.


pdeljanov:
I'm closing this PR since the `master` (v0.5.x branch) is effectively only in maintenance mode for small fixes. We can continue to track this in your `dev-0.6` targeted PR.

PR #415, opened 2025-11-04
symphonia-codec-aac: Additional HE-AAC detection in the decoder

mazza-av:
This change adds additional checks in the AAC decoder for Spectral Band Replication (SBR)  and Parametric Stereo (PS), which are not available in LC-AAC, in the Fill Element (ID_FIL).


pdeljanov:
Thanks, makes sense to me. If you can run this patch through rustfmt so the check passes that'd be great: `cargo +nightly fmt`. The clippy lints are fixed upstream.

PR #417, opened 2025-11-25
symphonia-format-isomp4: Adding short-circuits for well-formed DASH streams so we dont make too many http requests

phayes:
Hi there,

The symphonia-format-isomp4 wasn't treating DASH streams very well.  It was scanning the entire stream (causing many HTTP requests) before seeking back to the start.  

This PR fixes that by only reading what we need to read, allowing well-formed DASH streams to be initialized with minimal reading and seeking. 

My understanding of the MP4 format is fairly limited, so I would appreciate someone with some knowledge to review my logic and let me know if I made any bad assumptions.


phayes:
Oh I see a problem.  Marking as draft.

phayes:
Closing this - this wont work at all with other MP4 files that have a different layout.    Dang.

PR #419, opened 2025-12-18
Fix invalid channel count fuzzer crash

baylesj:
This patch fixes a crash where a FLAC buffer with an invalid channel count causes an abort. Instead, a proper decode error is returned.

See public fuzzer bug here:
http://crbug.com/468277952


baylesj:
Verified fixed in Chrome with the fuzzer. Please LMK if there's any additional coverage that would be helpful.

@pdeljanov, PTAL?

pdeljanov:
Thanks for finding and fix this!

baylesj:
Of course! Looking forward to upstreaming more fixes.

PR #420, opened 2025-12-18
Fix build issue due to manual rotate_left impl

baylesj:
Currently, Clippy builds are unhapy due to the manual implementation of `rotate_left()`:

```
error: there is no need to manually implement bit rotation
  --> symphonia-core/src/conv.rs:72:17
   |
72 |                 (x << k) | (x >> (32 - k))
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: this expression can be rewritten as: `x.rotate_left(k)`
   |
```

This patch removes the `rotl` method entirely in favor of using rotate_left() directly.

### DIFF
--- a/symphonia-core/src/conv.rs
+++ b/symphonia-core/src/conv.rs
@@ -67,16 +67,11 @@ pub mod dither {
                 }
             }
 
-            #[inline(always)]
-            fn rotl(x: u32, k: u32) -> u32 {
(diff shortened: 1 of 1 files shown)


baylesj:
@pdeljanov mind taking a look? Small change to fix the bots.

pdeljanov:
Thanks!

PR #421, opened 2025-12-26
Native dsd

ljufa:
Adds native dsd playback using alsa backend.
Related changes in cpal/alsa are submited here: https://github.com/RustAudio/cpal/pull/1078

All together is tested here: https://github.com/ljufa/rsplayer

### DIFF
--- a/symphonia-bundle-mp3/src/demuxer.rs
+++ b/symphonia-bundle-mp3/src/demuxer.rs
@@ -9,7 +9,7 @@ use symphonia_core::support_format;
 
 use symphonia_core::checksum::Crc16AnsiLe;
 use symphonia_core::codecs::CodecParameters;
-use symphonia_core::errors::{seek_error, unsupported_error, Result, SeekErrorKind};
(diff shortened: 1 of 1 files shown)


pdeljanov:
Very cool! DSD would be a very nice addition and was on my mind.

I've only skimmed this, but there are some issues before such a decoder could be merged:

1. All new features need to be based off the `dev-0.6` branch as `master` is only for the current `0.5.x` series which is  maintenance mode.
2. `AudioBuffer` is only for PCM audio, not for pass-through. To support this would require some additional changes to `symphonia-core`. That would need some thought, because right now any `GenericAudioBuffer` format can be converted to each other. Also, there are other formats beyond DSD that could benefit  from pass-through. So, until such cases are supported, any upstream DSD decoder would need to internally convert DSD to PCM.

These two issues seem far beyond your use-case, so I understand you probably won't be up for making these changes. You can always implement this in your application and register your own format reader and decoder into the format and codec registries.

ljufa:
Tnx for the feedback! In that case I will keep it on my fork/app untill proper implementation.

PR #422, opened 2025-12-26
Break infinite loop for corrupted radio stream

ljufa:
In my https://github.com/ljufa/rsplayer that I am using and testing daily, I've experienced infinite loop created by some online radio streams. (I can provide some examples if needed)

I guess this fix is not generic enough to solve corupted stream problems among all streams/codecs but in my case it helped. Open for ideas to solve it in more elegant way...


pdeljanov:
Ah, interesting problem.

Yeah, a format reader will attempt to re-sync infinitely until end-of-stream, but I suppose radio streams go on forever. `Probe` has a 1 MB limit, but there's nothing once a reader is instantiated. A generic solution would be to add a `FormatOption` indicating a limit (see `Limit` enum) on how many bytes to consume while attempting to re-sync. Then abide by that in the format readers.

Could you link some of those radio streams causing this? I just hope the re-sync logic is not too strict and exasperating the problem.

ljufa:
Atm, I could not find any stream that fails immediatelly and consistently, it happens randomly (might be after 1h). 
Anyway, current implementation without breaker is not robust enough for network streams. 
I can add logging to my app and check for patterns...

ljufa:
I found one url that fails and can be used to reproduce this issue `http://fr2.1mix.co.uk:9000/aach`

PR #425, opened 2026-01-04
feat(riff): Add RF64/BW64 format support for large WAV files

tphakala:
## Summary

This PR adds support for the RF64/BW64 file format, which extends the RIFF/WAV format to support files larger than 4GB.

**Motivation:** Autonomous Recording Units (ARUs) used in bioacoustics and environmental monitoring generate continuous audio recordings that frequently exceed the 4GB file size limit of standard WAV files. These devices output RF64-formatted files to accommodate the large data sizes. Without RF64 support, Symphonia cannot open these files.

**Changes:**
- Add `Ds64Chunk` parser for reading 64-bit size information
- Detect RF64 format marker and parse ds64 chunk when present
- Use 64-bit data size and sample count from ds64 for RF64 files
- Add RF64 format probe descriptor for automatic format detection
- Maintain full backward compatibility with standard WAV files

## Implementation Details

RF64 is specified in EBU Tech 3306 (MBWF / RF64: An extended File Format for Audio). Key aspects:

- RF64 files use `RF64` instead of `RIFF` as the stream marker
- A mandatory `ds64` chunk immediately after the header contains 64-bit sizes
- The 32-bit size fields are set to `0xFFFFFFFF` as placeholders
- For files < 4GB, standard WAV format continues to work unchanged

The implementation:
1. Detects RF64 vs RIFF marker in `try_new`
2. For RF64 files, parses the ds64 chunk to extract `data_size` and `sample_count`
3. Uses 64-bit sizes for RF64, falls back to 32-bit for standard WAV
4. Adds format probe to recognize RF64 files by magic bytes

## Testing

- Unit tests for ds64 chunk parsing (valid chunks, malformed chunks, chunks with table entries)
- Integration tests for RF64 format detection and reading
- Regression tests ensuring standard WAV files continue to work
- Manual testing with real ~4GB RF64 file from ARU device

## Test plan

- [x] `cargo test` passes
- [x] `cargo clippy` clean
- [x] `cargo +nightly fmt` applied
- [x] Manual testing with real RF64 file confirmed working

### DIFF
--- a/symphonia-format-riff/src/wave/chunks.rs
+++ b/symphonia-format-riff/src/wave/chunks.rs
@@ -566,8 +566,12 @@ pub struct Ds64Chunk {
 
 impl ParseChunk for Ds64Chunk {
     fn parse<B: ReadBytes>(reader: &mut B, _tag: [u8; 4], len: u32) -> Result<Self> {
-        // Minimum ds64 chunk size is 28 bytes (3 u64s + 1 u32 table length)
-        if len < 28 {
+        // ds64 fixed header: riffSize64 (8) + dataSize64 (8) + sampleCount64 (8) + tableLength (4)
+        const DS64_MIN_CHUNK_SIZE: u32 = 28;
+        // Each table entry: chunkId (4) + chunkSize64 (8)
+        const DS64_TABLE_ENTRY_SIZE: u64 = 12;
+
+        if len < DS64_MIN_CHUNK_SIZE {
             return decode_error("wav: malformed ds64 chunk");
         }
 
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks @tphakala for implementing this.

I'm happy to accept support for this format to Symphonia, however it'll need to be rebased for `dev-0.6` since `master`, the stable 0.5.x branch, is only for bug fixes at this time.

If you can point me in the direction of any publicly available test files as well, that would be helpful for me to add to my collection of test vectors.

pdeljanov:
Closing this one since the `dev-0.6` PR supersedes it.

PR #429, opened 2026-01-10
DRAFT: Feature/wavpack reader

dedobbin:
@pdeljanov As discussed here is the wavpack reader PR. I took a bit of a break since starting it, so i had to puzzle a bit finding the right commit where i cut out experiments with the decoder and other dev code.

It's very out of date because of the break, so first point of action on my side is resolving any conflicts/old things. Let me know what branch i should target

If you have any major concerns/suggestions about the structure we can go through it, but it's best for me to resolve any conflicts meanwhile, so the code is easier to review, hence i marked it draft

As far as i remember i marked branched from dev-0.6 but ill also double check this.


dividebysandwich:
I am very happy to see this PR! I have two questions though:
1) Does this support older v3 files as well?
2) Does it parse loop point data?

pdeljanov:
Hi @dedobbin, I skimmed over this draft and structurally it looks okay. You'll probably want to separate the sub-blocks into a separate module since there are many different types.

Some things things that will definitely need to change:

1. The `dev-0.6` branch switched to signed timestamps a few weeks ago, so if you rebase you will likely run into some build errors. This will also change your timestamp calculations a bit.
2. Since this is pretty far from a working WavPack demuxer, I think we should put it under an experimental feature flags: `exp-format-wavpack`.

dedobbin:
Hi @dividebysandwich 

This is just the reader part (obtaining the individual blocks/chunks, not decoding the data stream) so not much is done with loop points but my plan is to support them, as they are an interesting aspect of wavpack. But we will try and build the feature step for step.

As for older files, i need to dig a bit deeper into what you are exactly referring to, but if my memory serves right there are only extensions, so it should work then. If i misunderstand then let me know.


@pdeljanov 
Alright perfect. I'll start with these points. An experimental flag makes perfect sense to me.

dividebysandwich:
> Hi @dividebysandwich
> 
> This is just the reader part (obtaining the individual blocks/chunks, not decoding the data stream) so not much is done with loop points but my plan is to support them, as they are an interesting aspect of wavpack. But we will try and build the feature step for step.

That would be great! I'd need this for my RustyPipes project, there's a bunch of organ sample sets that use wavpack.
 
> As for older files, i need to dig a bit deeper into what you are exactly referring to, but if my memory serves right there are only extensions, so it should work then. If i misunderstand then let me know.

I am not really sure myself because the wavpack code isn't exactly straightforward and the version difference in the data isn't documented - but what I can tell you is that the command line tool doesn't support older versions of wavpack. Trying to unpack v3 files gives me a "This file version is no longer supported" error. 

ffmpeg can unpack these files but they have audio artifacts, it looks like ffmpeg introduces erroneous sample data at the end of each file.

PR #434, opened 2026-01-23
Remove sample clamping from MP3 and Vorbis decoders

### LINKED ISSUE #433: MP3 decoder: Option to disable sample clamping for peak detection use cases
## Summary

The MP3 decoder in `symphonia-bundle-mp3` clamps output samples to `[-1.0, 1.0]` in the synthesis stage. While this is appropriate for playback, it prevents accurate peak detection for audio analysis tools that need to detect clipping (samples exceeding 0 dBFS).

## Background

I'm developing [mp3rgain](https://github.com/M-Igashi/mp3rgain), a modern replacement for the classic mp3gain tool. It uses Symphonia for audio decoding to perform ReplayGain analysis and peak detection.

When analyzing MP3 files that would clip (peak > 1.0), the decoder reports a maximum of exactly 1.0, making it impossible to:
1. Detect files that are already clipping
2. Calculate accurate headroom for gain adjustments
3. Warn users about potential clipping after gain changes

Interestingly, the AAC decoder (`symphonia-codec-aac`) does NOT clamp samples, so M4A files correctly report peaks exceeding 1.0.

## Current Behavior

In `symphonia-bundle-mp3/src/synthesis.rs`, line ~180:

```rust
// Clamp and copy the PCM samples from o_vec to the output buffer.
let offset = b << 5;

for (o, s) in out[offset..offset + 32].iter_mut().zip(&o_vec) {
    *o = s.clamp(-1.0, 1.0);
}
```

## Example

Using mp3rgain with Symphonia:

```
# MP3 file - peak is clamped to 32768 (1.0 normalized)
$ mp3rgain -x clipping_audio.mp3
Max PCM sample: 32768.000000
(may be clipped - actual peak could be higher)

# Same audio as M4A - correct peak reported
$ mp3rgain -x clipping_audio.m4a  
Max PCM sample: 41531.960938
```

For comparison, the original mp3gain (using mpglib) reports the actual unclamped value:
```
$ mp3gain -x clipping_audio.mp3
Max PCM sample at current gain: 36772.700216
```

### Option 1: Add a decoder option to disable clamping

Add a field to `DecoderOptions` to control clamping behavior:

```rust
pub struct DecoderOptions {
    // ... existing fields ...
    
    /// If true, do not clamp output samples to [-1.0, 1.0].
    /// Useful for peak detection and audio analysis tools.
    /// Default: false (samples are clamped for safe playback)
    pub disable_sample_clamping: bool,
}
```

### Option 2: Remove clamping entirely (match AAC decoder behavior)

Since the AAC decoder doesn't clamp, the MP3 decoder could be made consistent. Applications that need clamping for playback can do so themselves.

### Option 3: Use a different output format

Output samples in a format that naturally supports out-of-range values (e.g., keep the pre-clamp f32 values available).

## Environment

- Symphonia version: 0.5.5
- Rust version: 1.83.0
- Platform: macOS (Apple Silicon)

## Related

- mp3rgain issue: https://github.com/M-Igashi/mp3rgain/issues/52
- mp3rgain PR with workaround: https://github.com/M-Igashi/mp3rgain/pull/53

I'd be happy to submit a PR if there's agreement on the preferred approach. Thank you for this excellent library!

M-Igashi:
## Summary

Remove the `[-1.0, 1.0]` sample clamping from MP3 and Vorbis decoders to align with FFmpeg behavior and enable accurate peak detection for audio analysis tools.

## Background

As discussed in #433:

- **FFmpeg (the de facto standard) does not clamp decoder output** - clipping only occurs when converting to integer formats
- The **AAC decoder in Symphonia already does not clamp**, so this change makes behavior consistent across codecs
- Clamping prevents accurate peak detection for audio analysis tools that need to detect clipping (samples exceeding 0 dBFS)

## Changes

| File | Change |
|------|--------|
| `symphonia-bundle-mp3/src/synthesis.rs` | Remove `clamp()` call, samples are now copied directly |
| `symphonia-codec-vorbis/src/dsp.rs` | Remove clamping loop entirely |

## Rationale

Applications that need clamped output for safe playback can easily apply clamping themselves after decoding:

```rust
for sample in buffer.iter_mut() {
    *sample = sample.clamp(-1.0, 1.0);
}
```

However, applications that need unclamped output (audio analysis, peak detection, ReplayGain calculation) cannot recover the original values if clamping is applied during decoding.

## Testing

- Built and tested with real MP3 and M4A files
- Verified decode-only playback works correctly with `symphonia-play`

Fixes #433

### DIFF
--- a/symphonia-bundle-mp3/src/synthesis.rs
+++ b/symphonia-bundle-mp3/src/synthesis.rs
@@ -322,11 +322,11 @@ pub fn synthesis(state: &mut SynthesisState, n_frames: usize, in_samples: &[f32]
             }
         }
 
-        // Clamp and copy the PCM samples from o_vec to the output buffer.
+        // Copy the PCM samples from o_vec to the output buffer.
         let offset = b << 5;
 
         for (o, s) in out[offset..offset + 32].iter_mut().zip(&o_vec) {
-            *o = s.clamp(-1.0, 1.0);
+            *o = *s;
         }
 
         // Shift the v_vec FIFO. The value v_front is the index of the 64 sample slot in v_vec
--- a/symphonia-codec-vorbis/src/dsp.rs
+++ b/symphonia-codec-vorbis/src/dsp.rs
@@ -131,10 +131,6 @@ impl DspChannel {
                 buf[self.bs0 / 2..].copy_from_slice(&self.imdct[end..self.bs1 / 2]);
             }
 
-            // Clamp the output samples.
-            for s in buf.iter_mut() {
-                *s = s.clamp(-1.0, 1.0);
-            }
         }
 
         // Save right-half of IMDCT buffer for later.


pdeljanov:
Thanks!

PR #436, opened 2026-01-26
isomp4: Gracefully handle malformed gnre atoms. (pdeljanov#311)

### LINKED ISSUE #311: "buffer underrun" when reading m4a file
Hi there, I'm using following code to read a file
```
    let file = Box::new(File::open(Path::new(filename)).unwrap());

    let mss = MediaSourceStream::new(file, Default::default());

    let hint = Hint::new();

    let format_opts: FormatOptions = Default::default();
    let metadata_opts: MetadataOptions = Default::default();

    let probed = symphonia::default::get_probe()
        .format(&hint, mss, &format_opts, &metadata_opts)             <--- error thrown this line
        .unwrap();

    probed.format
```

And got this error 
```
called `Result::unwrap()` on an `Err` value: IoError(Custom { kind: UnexpectedEof, error: "buffer underrun" })
```

After some research I found it should be thrown from `BufReader`, but I don't know what it is reading when throwing this error, please help take a look. I would probably also try to attach a debugger to see what's going on later maybe. I could also share the audio file but since it's purchased so I guess better share it privately.

Nich87:
## Summary

Gracefully handle malformed `gnre` atoms that contain less than 2 bytes of data.

## Problem

When reading certain M4A files, Symphonia throws a "buffer underrun" error during metadata parsing:

```
IoError(Custom { kind: UnexpectedEof, error: "buffer underrun" })
```

This occurs because some files contain malformed `gnre` (ID3v1 genre) atoms where the data is less than the expected 2 bytes. The current implementation unconditionally calls `read_be_u16()`, which fails when there isn't enough data.

## Solution

Instead of unconditionally reading the u16 value, this PR wraps the read operation in an `if let Ok(...)` pattern to gracefully skip malformed genre data. This ensures that:

1. Valid genre tags continue to work as expected
2. Malformed genre tags are silently ignored rather than causing a fatal error
3. The rest of the file's metadata and audio data remain accessible

## Changes

- Modified `add_id3v1_genre_tag()` in `symphonia-format-isomp4/src/atoms/ilst.rs` to gracefully handle read errors

```diff
     if let Some(value) = tag.values.first() {
-        // The ID3v1 genre is stored as a unsigned 16-bit big-endian integer.
-        let index = BufReader::new(&value.data).read_be_u16()?;
-
-        // The stored index uses 1-based indexing, but the ID3v1 genre list is 0-based.
-        if index > 0 {
-            if let Some(genre) = id3v1::util::genre_name((index - 1) as u8) {
-                builder.add_tag(Tag::new(Some(StandardTagKey::Genre), "", Value::from(*genre)));
+        // Gracefully handle malformed data (< 2 bytes). Fixes #311.
+        if let Ok(index) = BufReader::new(&value.data).read_be_u16() {
+            if index > 0 {
+                if let Some(genre) = id3v1::util::genre_name((index - 1) as u8) {
+                    builder.add_tag(Tag::new(Some(StandardTagKey::Genre), "", Value::from(*genre)));
+                }
             }
         }
     }
```

## Testing

- Verified that valid M4A files with proper genre tags still parse correctly
- Files with malformed genre atoms no longer cause a fatal error

Fixes #311.


pdeljanov:
Thanks!

PR #438, opened 2026-02-06
flac: Fix validation checks found by fuzzing

baylesj:
This commit adds validation checks to the FLAC decoder to prevent potential panics or invalid behavior when processing malformed streams.

- Verify frame channel count matches the stream information.
- Ensure dropped bits per sample does not exceed frame bits per sample.
- Validate that the predictor order does not exceed the block size.

Related Chromium bugs:

https://issues.chromium.org/issues/477312427
https://issues.chromium.org/issues/433557303

### DIFF
--- a/symphonia-bundle-flac/src/decoder.rs
+++ b/symphonia-bundle-flac/src/decoder.rs
@@ -120,6 +120,17 @@ impl FlacDecoder {
         self.buf.clear();
         self.buf.render_reserved(Some(header.block_num_samples as usize));
 
+        let frame_channels = match header.channel_assignment {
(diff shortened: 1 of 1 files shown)


pdeljanov:
Thanks, these changes all look very reasonable. I will port them to `dev-0.6` except for the channel validation fix since that one was already addressed in that branch.

baylesj:
Thank you for the quick review! I'm continuing work to replace FFMPEG with Symphonia as the audio decoding implementation in Chrome, so will continue pushing fuzzing bugs and other fixes / cleanups as that effort continues.

Issue #149, opened 2022-07-31
Add fuzzers


qarmin:
Looks that already some crates have code which can be used in fuzzers - https://github.com/pdeljanov/Symphonia/blob/master/symphonia-codec-aac/tests/tests.rs

If anyone want to implement it, this is git patch with unfinished attempt to do this - [symphonia.txt](https://github.com/pdeljanov/Symphonia/files/9228214/symphonia.txt)

Shnatsel:
Nice, thanks! The general framework looks workable. You said this is unfinished - what's left to be done?

I see the Vorbis target erroneously specifies the format as MP3, but... other than that :smiley: 

I'm actually aware of one item. Formats such as FLAC and OGG have checksums embedded. Checksum verification needs to be disabled when fuzzing, otherwise most inputs created by the fuzzer will be rejected because of the checksum mismatch, and no interesting decoding will happen.

If disabling the checksum is not exposed in the API, you can use `if cfg!(fuzzing) { return true }` in checksum validation functions. Fuzzers will set the `fuzzing` cfg flag automatically.

Shnatsel:
I've run the MP3 fuzzer for a bit and it didn't find any crashes. If my previous experience fuzzing multimedia libraries in C and Rust is anything to go by, this indicates that the fuzzing harness is broken.

I've checked, and line coverage for fuzzing MP3 looks reasonable: [mp3-coverage.html.gz](https://github.com/pdeljanov/Symphonia/files/9998262/mp3-coverage.html.gz)

@pdeljanov did you really manage to create an audio decoder that didn't crash when fuzzed for the first time?!

baylesj:
Looking at enhancing fuzzing support for version 0.6.

PR #443, opened 2026-02-10
Add CI workflow for fuzzing smoke tests

baylesj:
This patch adds a new step to the `check` job that ensures that the fuzzing manifest builds correctly.

This is necessary because the `run: cargo check --all --all-features --exclude symphonia-play` step does not build the fuzzer. As far as I can tell this is due to the `fuzz` directory defining its own `[workspace]` in `fuzz/Cargo.toml`. This effectively detaches it from the main project workspace, so running `cargo check --all` from the project root skips it.
      
Improvement on #149


pdeljanov:
A bit on the fence about this one since it would introduce non-determinism into the workflows. I think `cargo fuzz build` would be sufficient to check if the fuzz tests still are still building.

baylesj:
That's valid -- it would absolutely add a certain amount of non-determinism. I can add a fuzzing build target instead that just runs `cargo fuzz build` to make sure fuzzing is not completely broken, and then rely on OSS-Fuzz for actually executing fuzzing.

baylesj:
Can't seem to update the bug title for some reason, but I think this CL should be ready for another look.

pdeljanov:
This makes sense to me. Perhaps once significant time has been spent fuzzing Symphonia we could add the smoke tests.

PR #445, opened 2026-02-10
[Fuzzing] Fix fuzzing crashes found while adding support

baylesj:
This patch fixes a few fuzzing crashes found while adding fuzzing support -- one in the FLAC decoder, and one in the ALAC decoder.


baylesj:
@pdeljanov local fuzzing is already paying off and looking forward to getting coverage through OSS-Fuzz.

pdeljanov:
Thanks, these are nice finds.

Issue #439, opened 2026-02-06
Panic when probing broken mp3 file


ARoese:
When probing a specific invalid MP3 file with `enable_gapless` set to true, a panic occurs. I encountered this error when using rodio to play the mp3, but since the panic occurred in Symphonia, I made a minimum reproduction to post the issue here.

# Sample Code
```
use std::fs::File;
use symphonia::core::formats::FormatOptions;
use symphonia::core::io::MediaSourceStream;
use symphonia::default;

fn main() {
    let probe = default::get_probe();
    let mss = MediaSourceStream::new(
        Box::new(File::open("bad_mp3.mp3").expect("open audio file")),
        Default::default()
    );
    let pr = probe.format(&Default::default(), mss, &FormatOptions {
        enable_gapless:true,
        ..Default::default()
    }, &Default::default()).expect("probe failed");
}
```

Sample mp3:
[bad_mp3.mp3](https://github.com/user-attachments/files/25121494/bad_mp3.mp3)

# Unwind
```
thread 'main' (2861667) panicked at /home/atomr/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/symphonia-bundle-mp3-0.5.5/src/demuxer.rs:128:42:
attempt to subtract with overflow
stack backtrace:
   0:     0x6319d1137b02 - std::backtrace_rs::backtrace::libunwind::trace::hd57362190422746d
                               at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:     0x6319d1137b02 - std::backtrace_rs::backtrace::trace_unsynchronized::he61be86f5e8dfc7b
                               at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:     0x6319d1137b02 - std::sys::backtrace::_print_fmt::hd233ea33e70c29e2
                               at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/sys/backtrace.rs:66:9
   3:     0x6319d1137b02 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h718e2d17a1928e63
                               at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/sys/backtrace.rs:39:26
[...]

PR #450, opened 2026-02-18
mp3: ignore info tag if delay and padding exceed duration

liquid-dragons:
There are no checks that the delay and padding reported by the info tag make sense for the total duration, leading to the panic reported in #439.

However, since I'm new to Symphonia, I'm not sure what the best way would be to check for this. I chose to check when reading the info tag and error out, effectively ignoring the tag later on. Alternatively we could handle this while creating the `MpaReader`, and just ignore gapless if these fields don't make sense.

I've also checked the issue on 0.6, where this has been fixed already.


pdeljanov:
Hi @liquid-dragons,

Thanks for the PR. I think your alternative makes more sense. In the if-block on line 111, we add the delay and padding to the codec parameters and then pass them along to calculate other timing information.

I think this would be the appropriate place to check if they're valid, and simply ignore them and assume they're 0 for the remaining timing calculations. Otherwise, we'll lose the entire info tag with the current approach if the numbers are off.

So, the simplest solution would be to move that if-block into the if-let where we get the number of frames, then do your logic to ensure the padding and delay don't exceed the total number of frames. If they do, assume they're 0 instead and you can print a warning.

pdeljanov:
Thanks!

PR #451, opened 2026-02-19
[Fuzzing] Add component demuxer fuzz targets

baylesj:
This PR expands the fuzzing infrastructure for Symphonia to include new fuzzing target for each type of supported demuxer.

Continues work on #149


baylesj:
@pdeljanov PTAL when you have capacity.

pdeljanov:
LGTM, thanks!

PR #452, opened 2026-02-19
core: Replace buffer capacity panics with Result<()>

baylesj:
Refactors `AudioBuffer::convert`, `SampleBuffer::copy_*`, `RawSampleBuffer::copy_*`, and `Signal::render*` methods to return `Result<()>` instead of panicking when the destination buffer capacity is insufficient.

This improves robustness against malformed inputs or incorrect usage, ensuring that the library returns a handleable error (`decode_error("capacity will be exceeded")`) rather than crashing the application.

This change required updating call sites across the workspace to propagate or handle the new errors.

Tests have been added to `symphonia-core` to verify the new behavior.

We have previously worked on landing fuzzing fixes in this area for dev-0.6. Since a v0.5 maintainence release is imminent, I think these changes will help make that release more robust.

### DIFF
--- a/symphonia-play/src/output.rs
+++ b/symphonia-play/src/output.rs
@@ -104,6 +104,11 @@ mod pulseaudio {
             }
 
             // Interleave samples from the audio buffer into the sample buffer.
+            let n_samples = decoded.frames() * decoded.spec().channels.count();
+            if self.sample_buf.capacity() < n_samples {
+                error!("audio output write error: sample buffer capacity exceeded");
+                return Err(AudioOutputError::StreamClosedError);
+            }
             self.sample_buf.copy_interleaved_ref(decoded);
 
             // Write interleaved samples to PulseAudio.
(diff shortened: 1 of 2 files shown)


baylesj:
@pdeljanov PTAL when you have capacity.

pdeljanov:
Sorry, these types of changes can't be made on a 0.5.x branch because `symphonia-core` is a stable public interface that third-party out-of-tree decoders depend on so they can't be changed.

The best that could be done for 0.5.x is adding the necessary checks at the call sites, or being more strict about enforcing the pre-conditions that the decoders were seemingly written with in-mind.

If you want to rework this PR, you probably don't need as extensive changes. For example, the MPA family of decoders always have a large enough audio buffer to work with since the maximum capacity is always 1152 frames and the audio buffer is always initialized with that capacity in-mind.

baylesj:
Thank you for the review. In most of my work I don't have to worry too much about modifying stable public interfaces (and don't particularly miss worrying about ABI compatibilty). 

I'll refactor to produce a more conservative version.

PR #453, opened 2026-02-24
Complete `no_std` transition using `embedded-io` traits

tsar-boomba:
First of all, **huge** thank you to #412 for starting off this effort since this PR is based on their changes.

## Changes

This PR does large refactors to `symphonia-core` to replace the use of `std::io` traits with traits from [`embedded-io`](https://docs.rs/embedded-io/latest/embedded_io/). This is a breaking change to users, but a very minimal one, just requiring wrapping any `std` io objects in `FromStd` (provided by `symphonia-core`).

Besides that big change, I also added `hashbrown` to replace `std::collections::HashMap` and `HashSet` where used.

I also went through each crate, making them `no_std` and making sure they compile and pass tests.

## API Changes

Like I mentioned earlier the main change is the switch from `std` io traits to `embedded-io` traits. I chose `embedded-io` since its stable, and has the most support outside of `std`. A lot of the bigger changes are just copying functions and structs straight out of `std` that we can't access anymore (see `io::utils` in `symphonia-core`).

I also added an `std` feature to `symphonia-core` and `symphonia`. It's on by default in `symphonia`, but not `symphonia-core` which should allow for a pretty easy transition for users.

## Purpose

I'm working on a Digital Audio Player using Rust and would love to use Symphonia to support more audio codecs, so I decided to just get this done for my own project. I would really love to see this upstreamed, but if it doesn't work out I'll just maintain my own fork anyways. I'm also planning on adding async support for my own use case too.

I know its a lot, and I can try and split it up, but most of the consequential changes are in `symphonia-core`, so hopefully its not too much. Thank you for your hard work and time on this! I hope we can get this merged in.


tsar-boomba:
I'm closing this since I've finished the no_std + async transition in my fork and this PR is now outdated. Its working great for me on rp2350 with embassy.

oletf:
Hi, 
@tsar-boomba thanks a lot for the work on the no_std with embedded-io, it's very useful for my case !
(sorry for writing here, Issues aren't enabled on your fork)

I was wondering if the many `Send + Sync` (especially the ones related to `MediaSource`) constraints on various types could be gated behind a feature...
as in my `no_std` project, I use a file source which has to contain a `RefCell` and I don't have multithreading nor multiple cores on the microcontroller.

Issue #456, opened 2026-03-05
Symphonia's binary size footprint should be reduced


baylesj:
Filing this issue as a conversation around potential binary size improvements. For context, I am working on enabling Symphonia as the audio decoding solution on Android (see [Gerrit cl/7630806](https://chromium-review.git.corp.google.com/c/chromium/src/+/7630806?tab=comments)) and am analyzing the binary size footprint of Symphonia.

Adding FFMPEG for demuxing + audio resulted in ~660kb of binary size increase for Chrome on Android when added. Symphonia FLAC + MP3 support (not even all codecs, no demuxing yet) ends up in a 702kb binary size increase. I think there are some good opportunities to reduce this number, which makes including Symphonia on Chrome and in other embedded scenarios easier.

Some suggestions from resident video stack tech wizard Dale Curtis I intend to investigate:

1. Making the FFT_TWIDDLE_TABLE runtime generated instead of static, so it doesn't contribute to binary size.

2. create_audio_buffer() is apparently 25kb of code, this is a likely candidate for reduction.

3. Maybe not binary size related, but there are some mutable constants added apparently. Need to investigate this but adding here for posterity.

baylesj:
Looks like `lazy_static` just doesn't necessarily play well with the compiler, since the compiler will aggressively try to determine if something can be done at compile time.

I think in the dev-0.6 branch the right fix is probably to just stop using lazy_static now that [the standard library's std::sync::LazyLock type is preferred](https://github.com/rust-lang/rust-clippy/issues/12895).

pdeljanov:
Hey @baylesj,

These findings are quite surprising! You've put in a lot of effort, but before proceeding any further let's discuss a bit.

My expectation with `lazy_static` was that the memory for the static, pre-initialization, would be zero'd and reside in the .bss section and not occupy any space in the actual binary. However, I did my own test, and I see it's going into the .data section and bloating the binary. Very unexpected.

I'm also a bit surprised the twiddle tables are not being stripped out just in general because neither MP3 nor FLAC depend on the FFT. Anyways...

You've done quite a complex implementation with a `RwLock` and dynamically growing the table in the `master` branch, however I think that may be overly complex. Returning a Boxed array from the `lazy_static`, like you did in the `dev-0.6` branch is sufficient in my opinion and works fine even with `lazy_static`. In both cases, if a table isn't being used, the static will never be initialized and no additional heap memory will be used.

[...]

PR #457, opened 2026-03-06
core: Reduce binary bloat by using a dynamic FFT twiddle table

baylesj:
The previous implementation of the FFT module used a macro to generate 11
separate `lazy_static` arrays of precomputed twiddle factors for different
FFT sizes. This approach resulted in nearly 0.5 MiB of binary bloat.

This commit refactors the FFT module to use a single, dynamically resizable
`RwLock<Vec<Complex>>` for the twiddle factors.

- The table grows on demand up to the maximum required size.
- Smaller FFT operations reuse the larger table by accessing elements
  with a calculated stride, avoiding duplicate data.
- The hot inner loop was rewritten using zipped iterators to safely
  elide bounds checks without relying on `unsafe` code.
- Added comprehensive unit tests for striding, various sizes, and bounds.

This drastically reduces the compiled binary size and memory footprint
for applications that don't need the maximum FFT size, while preserving
the performance characteristics of the recursive Radix-2 algorithm.

#456

### DIFF
--- a/symphonia-core/src/dsp/fft.rs
+++ b/symphonia-core/src/dsp/fft.rs
@@ -713,25 +713,26 @@ mod tests {
 
     #[test]
     fn verify_fft_multithreaded() {
-        use std::thread;
         use std::sync::{Arc, Barrier};
+        use std::thread;
+        const NUM_THREADS: usize = 16;
         let mut handles = vec![];
-        let num_threads = 16;
-        let barrier = Arc::new(Barrier::new(num_threads));
+        let barrier = Arc::new(Barrier::new(NUM_THREADS));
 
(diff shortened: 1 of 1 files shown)


baylesj:
I'll work on fixing the failing check before requesting review.

baylesj:
@pdeljanov I believe this is ready for review, LMK what you think.

baylesj:
If you want me to move this to dev-0.6 or otherwise refactor I am happy to do so. The most streamlined version of this fix is to just put the static arrays in a Box.

pdeljanov:
Please see my general comment on #456.

baylesj:
Closing in favor of a more targeted approach.

PR #458, opened 2026-03-07
refactor: replace lazy_static with std::sync::LazyLock

baylesj:
This commit replaces the external `lazy_static` crate with the native `std::sync::LazyLock`, reducing external dependencies and leveraging the standard library for lazy initialization. Where applicable, static types are boxed in order to move their location from the static storage of the binary to the heap (where they should live, in my opinion, as lazy initialization for statically stored information computable at compile time does not seem to be particularly helpful).

This is a parallel but independent effort to #457 -- here the twiddle tables are updated to be heap allocated, but the more complex dynamic twiddle table in #457 is not included. If the dynamic twiddle table is approved, it can be merged into dev-0.6 after this patch lands.

* Removed the `lazy_static` dependency from all `Cargo.toml` files.
* Replaced `lazy_static! { ... }` macro blocks with `std::sync::LazyLock` declarations across the codebase (AAC, MP3, core, MKV, metadata, play, etc.).
* Formatted the converted blocks and preserved all original documentation comments.

NOTE: the core issue I am trying to solve, binary size bloat, doesn't necessary get fixed by this. But this is still a good lift for modern & idiomatic Rust in this library.

The major problem with lazy intialization is that it lets you think that your data does not get added to the binary size, but the truth is that if you introduce a type that the compiler can fold into static memory of the binary, the compiler will -- lazy lock or not. In that case lazy statics are actually worse than nothing since they take up all of the space in the binary of a normal static, but the compiler has a harder time optimizing them out as unused and you still end up potentially calculating the values at runtime (meaning your binary is bloated by an uninitialized array). I'm frustrated & surprised by this enough that I think I shall write a post of some kind about.

Helps #456.

### DIFF
--- a/symphonia-codec-aac/src/aac/ics/mod.rs
+++ b/symphonia-codec-aac/src/aac/ics/mod.rs
@@ -591,4 +591,3 @@ fn read_escape<B: ReadBitsLtr>(bs: &mut B) -> Result<u16> {
 
     Ok(word)
 }
-


baylesj:
@pdeljanov, if we can land #457 I'm happy to merge that patch into this one for the dev-0.6 branch.

LMK what you think.

pdeljanov:
Please see my general comment on #456.

baylesj:
Closing in favor of a more targeted fix.

PR #459, opened 2026-03-08
[ISOMP4] Fix no samples found demuxing HLS stream

tsar-boomba:
The current logic in `try_read_more_segments` will stop too early if the stream is segmented because it still needs to read the next one. This PR just fixes that by only ending early if the stream is not segmented.


tsar-boomba:
I also added duration for seekable streams by summing the frames in each sidx atom found

baylesj:
Tests?

pdeljanov:
Thanks for your PR. I'm closing this since this should be fixed, plus a bit more, in ad5f2d7f117523bdb8e18c3bf7e23fc933afa946 as part of 0.6.0.

PR #478, opened 2026-04-08
vorbis: Clamp Floor1 values to prevent out-of-bounds access

baylesj:
This patch addresses a crash in the Vorbis decoder by implementing bounds checking and value clamping in the Floor1 implementation. 

Instead of clamping values inside the inner loop of `render_line` (which impacts performance), this change applies clamping once per line segment in `synthesis_step2`. This guarantees the Y values will map correctly to the 256-element inverse dB table.

Additionally, the `render_line` function was hardened with early exits and explicit checks to ensure that both the starting and ending X coordinates remain within the current block size. These changes prevent malformed or malicious bitstream data from triggering out-of-bounds Rust panics and subsequent application aborts.

Issue discovered by fuzzing in Chrome, see https://crbug.com/498549512.

### DIFF
--- a/symphonia-codec-vorbis/src/floor.rs
+++ b/symphonia-codec-vorbis/src/floor.rs
@@ -632,12 +632,12 @@ impl Floor1 {
         let mut hx = 0;
         let mut hy = 0;
         let mut lx = 0;
-        let mut ly = floor_final_y0 * multiplier;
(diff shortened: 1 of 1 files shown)


baylesj:
@pdeljanov I think this is now in a pretty solid state for review. PTAL.

pdeljanov:
LGTM, thank you!

Issue #480, opened 2026-04-18
Certain ISO-MP4 files have their audio track decoded as junk


marin-m:
See https://github.com/marin-m/SongRec/issues/263 for a sample test file

Also, when I just disable the MP4 decoding backend in the Rodio create features it will still try to incorrectly decoding using the MP3 then the WAV decoder (??), I need to disable these too to have Rodio try other backends.

Perhaps related to #475 / #332?

Regards,

sscobici:
The provided sample is HE-AAC, which Symphonia does not yet support. See PR #473 for HE-AAC and HE-AAC v2 support.

pdeljanov:
The decoded audio is not junk. HE-AAC/HE-AACv2 is a form of progressive enhancement. Without a decoder that supports it, it decodes at half the sample rate.

As for the probing issue, that's an chronic issue that's difficult to solve. As you de-register format readers, the probe starts to get lobotomised. If the MP4 has a byte sequence that matches that of a MP3 header, then the MP3 reader can get chosen. The `v0.6.0-alpha.2` version of Symphonia should improve this, however, Rodio is stuck on the stable `v0.5.5` release until the next version is released, and they update to the newer APIs.

PR #473, opened 2026-03-24
feat(aac): Support to HE-AAC and HE-AAC v2

felipehertzer:
Hey,

I have added support to HE-AAC and HE-AAC v2.

You can get file samples [here](https://www2.iis.fraunhofer.de/AAC/stereo.html).

This implementation follows [ISO/IEC 14496-3](https://ossrs.io/lts/zh-cn/assets/files/ISO_IEC_14496-3-AAC-2001-7f4d0b3622b322cb72c78f85d91c449f.pdf)

Codec | Extra tools | Sweet spot
-- | -- | --
AAC-LC | — | 96–320 kbps
HE-AAC v1 | SBR | 48–80 kbps
HE-AAC v2 | SBR + PS | 24–40 kbps

Thank you.

-------
obs: In this implementation was used AI (Claude).

PR #481, opened 2026-04-18
mp4: add audio profile detection

sscobici:
HE-AAC and HE-AAC v2 are not detected correctly, and Symphonia attempts to decode them as LC.

Fix SBR and PS detection according to ISO/IEC 14496-3 (section 1.6.2.1, AudioSpecificConfig).

Expose the profile for MP4.

This will provide a clearer error explanation for #480 instead of attempting to decode it.

### DIFF
--- a/symphonia-common/src/mpeg/audio/mod.rs
+++ b/symphonia-common/src/mpeg/audio/mod.rs
@@ -6,6 +6,8 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use symphonia_core::audio::{Channels, layouts};
+use symphonia_core::codecs::CodecProfile;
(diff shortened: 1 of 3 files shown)


pdeljanov:
Thanks for the fix. I think it may be worthwhile to allow the AAC decoder to still play HE-AAC/HE-AACv2 at half the sampling rate since this is a form of progressive enhancement. Any issues with that?

sscobici:
I’m not sure if this will solve the track detection issue reported in https://github.com/marin-m/SongRec/issues/263
Also, symphonia-check produces many failed samples when playing at half the sampling rate.

Otherwise, I don't have issues with this, at least until the PR #473 for HE-AAC and HE-AAC v2 support is reviewed and merged.

PR #493, opened 2026-04-20
isomp4: improvements

sscobici:
- add detection of AC-4
- populate track with language
- increased max size for video extra data
- check for max allowed atom size to avoid overflow in debug, see attachment

https://github.com/user-attachments/assets/7b474105-c78f-47e7-a681-e1d2ee2cda05

### DIFF
--- a/symphonia-format-isomp4/src/atoms/avcc.rs
+++ b/symphonia-format-isomp4/src/atoms/avcc.rs
@@ -25,14 +25,14 @@ pub struct AvcCAtom {
 impl Atom for AvcCAtom {
     fn read<R: ReadAtom>(it: &mut AtomIterator<R>, header: &AtomHeader) -> Result<Self> {
         // TODO: Validate.
-        const MAX_AVCC_ATOM_SIZE: u64 = 1 * 1024;
+        const MAX_AVCC_ATOM_SIZE: u64 = 4 * 1024;
 
         // The AVCConfiguration atom payload is a single AVCDecoderConfigurationRecord. This record
         // forms the defacto codec extra data. It should not exceed 1 kB.
         let len = match header.data_size() {
             Some(len) if len <= MAX_AVCC_ATOM_SIZE => len as usize,
             Some(_) => {
-                return decode_error("isomp4 (avcC): atom size is greater than 1 kb");
+                return decode_error("isomp4 (avcC): atom size is greater than 4 kb");
             }
             None => {
                 return decode_error("isomp4 (avcC): expected atom size to be known");
--- a/symphonia-format-isomp4/src/atoms/dac4.rs
+++ b/symphonia-format-isomp4/src/atoms/dac4.rs
@@ -1,0 +1,42 @@
+// Symphonia
+// Copyright (c) 2019-2022 The Project Symphonia Developers.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use symphonia_core::codecs::audio::well_known::CODEC_ID_AC4;
+
+use crate::atoms::stsd::AudioSampleEntry;
+use crate::atoms::{Atom, AtomHeader, AtomIterator, ReadAtom, Result, decode_error};
+
+#[derive(Debug)]
+pub struct Dac4Atom {
+    /// AC4SpecificBox
+    extra_data: Box<[u8]>,
+}
+
+impl Atom for Dac4Atom {
+    fn read<R: ReadAtom>(it: &mut AtomIterator<R>, header: &AtomHeader) -> Result<Self> {
+        // TODO: Validate.
+        const MAX_DAC4_ATOM_SIZE: u64 = 4 * 1024;
+
+        // AC4SpecificBox should have length
+        let len = match header.data_size() {
+            Some(len) if len <= MAX_DAC4_ATOM_SIZE => len as usize,
+            Some(_) => return decode_error("isomp4 (dac4): atom size is greater than 4 kb"),
+            None => return decode_error("isomp4 (dac4): expected atom size to be known"),
+        };
+
+        let extra_data = it.read_boxed_slice_exact(len)?;
+
+        Ok(Dac4Atom { extra_data })
+    }
+}
+
+impl Dac4Atom {
+    pub fn fill_audio_sample_entry(&self, entry: &mut AudioSampleEntry) {
+        entry.codec_id = CODEC_ID_AC4;
+        entry.extra_data = Some(self.extra_data.clone());
+    }
+}
--- a/symphonia-format-isomp4/src/atoms/ftyp.rs
+++ b/symphonia-format-isomp4/src/atoms/ftyp.rs
@@ -33,7 +33,12 @@ impl Atom for FtypAtom {
         }
 
         // Major
-        let major = FourCc::new(it.read_quad_bytes()?);
+        let Some(major) = FourCc::try_new(it.read_quad_bytes()?)
+        else {
+            return decode_error(
+                "isomp4 (ftyp): major - only ASCII characters are allowed in a FourCc",
+            );
+        };
 
         // Minor
         let minor = it.read_quad_bytes()?;
@@ -46,8 +51,13 @@ impl Atom for FtypAtom {
         let mut compatible = Vec::with_capacity(MAX_TABLE_INITIAL_CAPACITY.min(n_brands as usize));
 
         for _ in 0..n_brands {
-            let brand = it.read_quad_bytes()?;
-            compatible.push(FourCc::new(brand));
+            let Some(brand) = FourCc::try_new(it.read_quad_bytes()?)
+            else {
+                return decode_error(
+                    "isomp4 (ftyp): brand - only ASCII characters are allowed in a FourCc",
+                );
+            };
+            compatible.push(brand);
         }
 
         Ok(FtypAtom { major, minor, compatible })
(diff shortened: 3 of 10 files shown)


symphonia-format-isomp4/src/atoms/avcc.rs
```diff
@@ -25,14 +25,14 @@ pub struct AvcCAtom {
 impl Atom for AvcCAtom {
     fn read<R: ReadAtom>(it: &mut AtomIterator<R>, header: &AtomHeader) -> Result<Self> {
         // TODO: Validate.
```
baylesj, line 27: Can you update this TODO as well? Where does this max atom size come from?

symphonia-format-isomp4/src/atoms/dac4.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Debug)]
+pub struct Dac4Atom {
+    /// AC4SpecificBox
+    extra_data: Box<[u8]>,
+}
+
+impl Atom for Dac4Atom {
+    fn read<R: ReadAtom>(it: &mut AtomIterator<R>, header: &AtomHeader) -> Result<Self> {
+        // TODO: Validate.
```
baylesj, line 21: Same here as above -- where does it come from and can you update the comment?

symphonia-format-isomp4/src/atoms/dac4.rs
```diff
@@ -0,0 +1,42 @@
...
+        const MAX_DAC4_ATOM_SIZE: u64 = 4 * 1024;
+
+        // AC4SpecificBox should have length
+        let len = match header.data_size() {
+            Some(len) if len <= MAX_DAC4_ATOM_SIZE => len as usize,
+            Some(_) => return decode_error("isomp4 (dac4): atom size is greater than 4 kb"),
```
baylesj, line 27: Nit: decode error should not reference the actual MAX_DAC4_ATOM_SIZE. to maintain a string literal here I would just say `atom size exceeded maximum value` or similar.

symphonia-format-isomp4/src/atoms/dac4.rs
```diff
@@ -0,0 +1,42 @@
...
+            None => return decode_error("isomp4 (dac4): expected atom size to be known"),
+        };
+
+        let extra_data = it.read_boxed_slice_exact(len)?;
+
```
baylesj, line 32: Nit: unnecessary space here.

symphonia-format-isomp4/src/atoms/hvcc.rs
```diff
@@ -25,13 +25,14 @@ pub struct HvcCAtom {
 
 impl Atom for HvcCAtom {
     fn read<R: ReadAtom>(it: &mut AtomIterator<R>, header: &AtomHeader) -> Result<Self> {
-        const MAX_HVCC_ATOM_SIZE: u64 = 1 * 1024;
+        // TODO: Validate.
```
baylesj, line 28: Same as above. TODO comment is not helpful here IMO.

symphonia-format-isomp4/src/atoms/avcc.rs
sscobici: According to ISO_IEC_14496-15, 5.3.3.1.2 theoretical limit of AVCDecoderConfigurationRecord is about 32 * 32Mb ~ 1Gb.
I'm not sure there is a need to allow so much, but agree that 4k might be too low.
Overall symphonia needs improvements on controlling the max memory needed for demuxer. 
@pdeljanov any high level suggestions for limits?
I know ffmpeg have a set of limits for different parts but not sure if we want to apply the same.

symphonia-format-isomp4/src/atoms/stsc.rs
```diff
@@ -51,6 +53,50 @@ impl StscAtom {
...
+            let mut total_chunks = u32::MAX;
+
+            if let Some(stco) = stco {
+                total_chunks = stco.chunk_offsets.len() as u32;
+            }
+            else if let Some(co64) = co64 {
+                total_chunks = co64.chunk_offsets.len() as u32;
+            }
+
+            for i in 0..self.entries.len() - 1 {
```
baylesj: The validation loop `for i in 0..self.entries.len() - 1 `skips the `first_chunk >= total_chunks` bounds check entirely if the stsc table has exactly 1 entry (because `len - 1` evaluates to 0,
  making the loop skip execution). Single-entry stsc tables are extremely common in standard MP4 files. 

Suggestion: Check the last entry's bounds before the loop. Because chunks are strictly monotonic, validating the last entry bounds-checks the entire sequence.

symphonia-format-isomp4/src/atoms/stsc.rs
```diff
@@ -51,6 +53,50 @@ impl StscAtom {
...
+                if self.entries[i].first_chunk >= total_chunks
+                    || self.entries[i + 1].first_chunk >= total_chunks
+                {
+                    return decode_error(
+                        "isomp4 (stsc): entry's first chunk is bigger than total chunk number",
+                    );
+                }
+
+                // Validate that first_chunk is monotonic across all entries.
+                if self.entries[i + 1].first_chunk <= self.entries[i].first_chunk {
```
baylesj: While redundant entries are inefficient, some poorly-behaved encoders might emit consecutive runs with the same first_chunk (effectively representing a 0-length run where n = 0). The old
  implementation handled this mathematically fine. Rejecting these now might break files that previously played without issue. Consider reverting this to < unless explicitly intended to reject
  redundant runs.

symphonia-format-isomp4/src/atoms/stsc.rs
```diff
@@ -64,37 +110,26 @@ impl Atom for StscAtom {
...
-                    return decode_error("isomp4 (stsc): entry has 0 samples per chunk");
-                }
-
-                let n = entries[i + 1].first_chunk - entries[i].first_chunk;
-
-                entries[i + 1].first_sample =
-                    entries[i].first_sample + (n * entries[i].samples_per_chunk);
+            // Validate that first chunk is indexed from 1
+            if first_chunk == 0 {
+                return decode_error("isomp4 (stsc): entry's first chunk undex should be from 1");
```
baylesj: Nit: typo

symphonia-format-isomp4/src/atoms/ftyp.rs
```diff
@@ -33,7 +33,12 @@ impl Atom for FtypAtom {
         }
 
         // Major
-        let major = FourCc::new(it.read_quad_bytes()?);
+        let Some(major) = FourCc::try_new(it.read_quad_bytes()?)
+        else {
+            return decode_error(
+                "isomp4 (ftype): major - only ASCII characters are allowed in a FourCc",
```
baylesj: ftyp not ftype right?

symphonia-format-isomp4/src/atoms/hdlr.rs
```diff
@@ -56,7 +55,7 @@ impl Atom for HdlrAtom {
             b"subt" => HandlerType::Subtitle,
             b"text" => HandlerType::Text,
             &hdlr => {
-                warn!("unknown handler type {:?}", FourCc::new(hdlr));
+                warn!("unknown handler type {:?}", hdlr);
```
baylesj, line 58: Consider wrapping `hdlr` still to make sure it's not outputted as a raw byte array. Maybe `String::from_utf8_lossy(&hdlr)`?

baylesj (commented):
Given the changes to the parser constraints, it would be highly beneficial to add:
   1. An integration test with a small ISOMP4 file containing an AC-4 audio track to ensure the new dac4 atom is parsed and correctly mapped to CODEC_ID_AC4.
   2. A unit test or fuzzer case for the bounds check in stsc.rs to ensure files with malformed first_chunk boundaries fail to decode gracefully without panicking.

symphonia-format-isomp4/src/atoms/dac4.rs
sscobici: Do you mean empty lines? I think it aligns with the style in other files.

symphonia-format-isomp4/src/atoms/hdlr.rs
sscobici: This will not print non utf8 chars.
Here is the example output for "vide" where 'd' was replaced with 0x05 which is not a printable char
`WARN  symphonia_format_isomp4::atoms::hdlr > unknown handler type [118, 105, 5, 101]`

I think in this case we want to see a raw byte array.

symphonia-format-isomp4/src/atoms/ftyp.rs
sscobici: thanks

symphonia-format-isomp4/src/atoms/stsc.rs
sscobici: thanks

symphonia-format-isomp4/src/atoms/stsc.rs
sscobici: Thanks,

According to ISO spec C.7.1 Data Location
> Samples are, in fact, stored in contiguous runs of samples for one track; these runs are called chunks.

I believe the first_chunk must be strictly monotonic.
But as you mentioned, having redundant runs will not break the playback, so I adjusted the check to validate for monotonic as it was in original code.

symphonia-format-isomp4/src/atoms/stsc.rs
sscobici: Thanks for suggestion, adjusted

sscobici:
closing as crash fixes were applied to master, others are minor and probably not mature for merging.
Language should be normalized accross containers for example,

PR #494, opened 2026-05-01
mkv: set track duration

sscobici:
Sets duration for mkv. Improved version for mkv vs the simple one from #332.
Takes duration from tag if it exists.

Feel free to change the implementation and add DURATION into some standard tag, if you want.


pdeljanov:
Hey @sscobici , sorry to say, but I *really* do not like using user-metadata for setting the duration. I've avoided that even for MP3 where the `TLEN` ID3v2 frame is used for that.

I took a look at ffmpeg's behaviour and I see that they in fact do not populate the duration for the track. But, of course, they have a media duration field. So I added a media info. struct in 3d612dd0d79f18cc013f3c01ab33e77745011b58, and the MKV implementation populating it is in 8c2e951b8d06fa8b6f7d0ac14abd9127384995d9 (along with improved timestamp handling).

I also populated the track and media durations for everything other than MP4 since you were grinding away at that in the other PR.

Take a look and see what you think. I'm going to tentatively close this PR, but feel free to reopen if you think something else needs to be addressed.

Issue #166, opened 2022-12-04
Is possible that decoder support PacketRef?


simon-fu:
Suppose application has a fixed buffer to store every packet data, but before invoking Decoder::decode()  it must invoke Packet::new() to  alloc data first.  That make poor performance. Is possible that add a new type PacketRef and change decode function  signature to Decoder::decode(&PacketRef) ?
The PacketRef look like:
``
pub PacketRef<'a> {   
  .. the same as Packet,   
  data: &'a [u8],   
}   
``

pdeljanov:
Hey @simon-fu,

I've run into the same problem too when trying to do a zero-copy decode. We could likely improve this in v0.6 since we would be making breaking changes to the API.

That being said, have you found that to be a major performance problem? Packets are generally small so the allocation should be pretty fast. 

---

My previous ideas:

1. `Packet` could become an enum of either a borrowed slice or owned Box, but that might introduce tricky lifetime problems.
2. Have a second `decode` function that takes a `&[u8]` slice or `PacketRef` as you described.

pdeljanov:
> Just wanted to note that this seems related: #109

Yep, a similar concern, but just at the output side.

Thanks for your patience regarding that PR. I'm fairly confident we can address both issues in v0.6, but I just want to get a bunch of quality and performance fixes out in v0.5.2 first.

I have many new thoughts to share on the relationship between `AudioBuffer`s and `SampleBuffer`s that hopefully we can use to further improve the API.

pdeljanov:
Implemented in 929bdc0d3afc4de285cd2d76a7e423828c7260cc.

PR #495, opened 2026-05-05
Introduce PacketRef for zero-copy packet passing

baylesj:
Resolves issue #166.

This change introduces a new `PacketRef` structure that acts as a borrowed version of `Packet`, allowing for zero-copy decoding when application data already resides in a fixed buffer or when data is passed from an external FFI environment (e.g., C++ std::vector) as a slice.

- Implemented `PacketRef` struct with slice borrowing
- Updated AudioDecoder and SubtitleDecoder traits to accept PacketRef
- Adapted all decoders (AAC, ALAC, ADPCM, FLAC, MP3, PCM, Vorbis)
- Updated tests, examples, and fuzzers
- Added PacketRef specific unit tests in packet.rs

I verified these changes programmatically using a custom GlobalAlloc implementation which tracked memory allocations. The test proved that constructing a PacketRef from a pre-allocated slice eliminated the heap allocation and data copying that Packet previously required.

### DIFF
--- a/symphonia-bundle-flac/src/decoder.rs
+++ b/symphonia-bundle-flac/src/decoder.rs
@@ -259,7 +259,7 @@ impl AudioDecoder for FlacDecoder {
         &self.params
     }
 
-    fn decode_ext(&mut self, packet: &PacketRef<'_>) -> Result<GenericAudioBufferRef<'_>> {
+    fn decode_ref(&mut self, packet: &PacketRef<'_>) -> Result<GenericAudioBufferRef<'_>> {
         if let Err(e) = self.decode_inner(packet) {
             self.buf.clear();
             Err(e)
--- a/symphonia-bundle-flac/src/decoder.rs
+++ b/symphonia-bundle-flac/src/decoder.rs
@@ -20,7 +20,7 @@ use symphonia_core::codecs::audio::{AudioDecoder, FinalizeResult, VerificationCh
 use symphonia_core::codecs::registry::{RegisterableAudioDecoder, SupportedAudioCodec};
 use symphonia_core::errors::{Error, Result, decode_error, unsupported_error};
 use symphonia_core::io::{BitReaderLtr, BufReader, ReadBitsLtr};
-use symphonia_core::packet::Packet;
+use symphonia_core::packet::PacketRef;
 use symphonia_core::support_audio_codec;
 use symphonia_core::util::bits::sign_extend_leq32_to_i32;
 
@@ -135,7 +135,7 @@ impl FlacDecoder {
         })
     }
 
-    fn decode_inner(&mut self, packet: &Packet) -> Result<()> {
+    fn decode_inner(&mut self, packet: &PacketRef<'_>) -> Result<()> {
         let mut reader = packet.as_buf_reader();
 
         // Synchronize to a frame and get the synchronization code.
@@ -259,7 +259,7 @@ impl AudioDecoder for FlacDecoder {
         &self.params
     }
 
-    fn decode(&mut self, packet: &Packet) -> Result<GenericAudioBufferRef<'_>> {
+    fn decode_ext(&mut self, packet: &PacketRef<'_>) -> Result<GenericAudioBufferRef<'_>> {
         if let Err(e) = self.decode_inner(packet) {
             self.buf.clear();
             Err(e)
--- a/symphonia-bundle-mp3/src/decoder.rs
+++ b/symphonia-bundle-mp3/src/decoder.rs
@@ -13,7 +13,7 @@ use symphonia_core::codecs::audio::{
 use symphonia_core::codecs::registry::{RegisterableAudioDecoder, SupportedAudioCodec};
 use symphonia_core::errors::{Result, decode_error, unsupported_error};
 use symphonia_core::io::FiniteStream;
-use symphonia_core::packet::Packet;
+use symphonia_core::packet::PacketRef;
 use symphonia_core::support_audio_codec;
 
 #[cfg(feature = "mp1")]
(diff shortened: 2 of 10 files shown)


baylesj:
@pdeljanov had a surprising amount of clippy errors, but I got the bots happy. LMK if there's anything else you would like to see in this patch set -- would be nice to eliminate an unnecessary copy for Symphonia, both in and outside of its Chromium usage. I just successfully rolled Chromium to use the alpha.2 release and am interested in continuing to contribute to the ongoing dev-0.6 work.

pdeljanov:
@baylesj , Looks like you're targeting the wrong branch. Development has moved to `main` (the new default branch where the 0.6 alphas have also been cut from). So, `dev-0.6` is quite stale at this point. There will be no clippy warnings on `main`. More details here: https://github.com/pdeljanov/Symphonia/issues/330#issuecomment-4115283252.

Can you rebase and re-target?

I'll do a more thorough review at that point. However, I think my main request for this PR will be to keep the `decode` function as-is (i.e., taking a `&Packet`). You can add a new required function to the trait, `decode_ref` (open to other suggestions) that takes the `&PacketRef<'_>`. Change the decoders to use this. Then, provide a default implementation for `decode` function that simply calls `self.decode_ref(packet.as_packet_ref())`. In this way we can keep things simple for the primary use-case of demuxing and decoding with Symphonia.

baylesj:
@pdeljanov: Rebased and renamed the methods as suggested // keeping the use case of demuxing + decoding with Symphonia simple sounds great to me. Long term it would be awesome to use Symphonia for demuxing in Chrome as well, assuming Symphonia is interested in and can get to the point where it properly handles all of the video related container stuff as well (which I don't know if that would actually be a goal or not).

It's great that we are back working on the `main` branch again! I'm using the latest alpha release in Chrome and it's been good so far!!

pdeljanov:
@baylesj , thanks. I will follow this up with some additional changes around `Packet` and `PacketRef`. Mainly to integrate `PacketRef` with `PacketBuilder` and also do some cleanups on `Packet` that should be propagated to `PacketRef`.

---
Happy to hear the Chrome integration efforts are going well! Which codecs are you using now? Any insights on Symphonia vs. ffmpeg you're seeing (i.e., speed, support, etc.)?

The long-term goal of Symphonia now is to do the entire media demuxing and decoding process. Basically, it should be possible to use Symphonia to create a proper video player. That being said, as much as I'm personally interested in doing a video decoder, it's probably not in the cards for a long time. I think I would wrap a crate like rav1d, or implement a HW video decoder to bring-up video support.

My immediate goals post 0.6.0 release are: 1) Implementing an Opus decoder, 2) Finishing demuxer support for video (e.g., populate DTS, etc.), and 3) Getting a proof-of-concept video decoder going using a HW decoder or rav1d.

I think the big challenge to using Symphonia for demuxing in a browser would be support for MSE? Admittedly, I've not looked into how this is implemented with ffmpeg either. Async is a popular feature request, so I'd like any future redesign of `FormatReader` to be able to support all three use-cases: sync, async, and MSE. Presumably the majority of a container's implementation could be shared and used by thin shims for each use-case.

I'm open to supporting this in the future, but on a practical level I would need help in understanding the requirements before coming up with a design, and also substantial assisstance refactoring the existing readers to support whatever the new design is.

baylesj:
I'm currently experimenting with enabling Symphonia for FLAC, MP3, Vorbis, and PCM decoding, with the goal of removing the FFmpegAudioDecoder entirely. The codec gaps for Symphonia are Opus (which I am working around with a C++ OpusAudioDecoder based on libopus) and HE-AAC (which I'm working with the Fraunhofer and Android folks to figure out a good resolution for). 

At this point Symphonia integration is in a stabilization phase -- fixing security issues, improving the implementation by eliminating extra copies and doing performance profiling, and fixing decode errors. I have seen an increase in decoding errors and a significant regression in performance over FFmpeg, but the gap continues to close with both the upstream work you and I have both been doing as well as improvements I've been making in Chromium's integration.

Container wise, Symphonia wouldn't actually have to handle MSE in order to be viable for demuxing in Chrome -- it would just need to replace the FFmpegDemuxer (which does not handle MSE or HLS) for MP4, Ogg, WebM, WAV, and Matroshka. Migrating the ChunkDemuxer (which handles MSE) to Rust could easily be a separate project. I think Symphonia already has "good" level support for all of the containers we would need, we would just want to get that support to "excellent."

As far as video codecs go, there is a lot of interest in Chrome around moving off of FFmpeg (especially as of late since we have been very understandably going through a massive influx of security bugs) and my understanding is that we currently only use it for H264 at this point. Migrating other codecs such as VP9 or AV1 to Rust also has interest for us and we are currently trying to evaluate the best path forward.

My hope is to find a good path forward where we can make some significant security wins for Chrome and also give back to the open source Rust community in a symbiotic relationship, the same way I have been trying to approach the Symphonia audio decoder migration. I am  interested in exploring Symphonia for demuxing and also seeing what the future holds for video decoding.

pdeljanov:
Sounds like things have been moving along!

I'm curious about the performance regressions you are seeing. Is that on 0.5.5 without `opt-simd`? Using the benchmarking methodology documented in the repo. and 0.6.0 with `opt-simd`, I usually see a +/-15% difference using a random sampling of my personal music library. Of course, that's a small sample size, but I'm wondering what's being missed if there is a "significant regression".

Feel free to open an issue if there you think there is something particularly bad.

> Container wise, Symphonia wouldn't actually have to handle MSE in order to be viable for demuxing in Chrome -- it would just need to replace the FFmpegDemuxer (which does not handle MSE or HLS) for MP4, Ogg, WebM, WAV, and Matroshka. 

Oh good! I suspect the format readers will need quite a bit more work than the decoders, but they're slowly getting there now that there is more focus on getting them workable for more than just audio.

Thanks for the update!

baylesj:
I've been seeing a roughly 30% performance decrease in comparison to FFmpeg on 0.6.0, with a minor +/- 5% difference when I enable `opt-simd`. This is of course in C++ benchmarking tests and comparing a C/C++ FFmpeg-only with a FFmpegDemuxer + SymphoniaAudioDecoder using Symphonia through an FFI layer.

I need to do additional investigating around what the performance bottlenecks are before I can say anything as definitive as Symphonia is slower -- it could be FFI-related, something about my specific implementation like the copying this patch is removing, or any number of oddities.

Once I have a better picture I'll report back and potentially file issues or patches.

PR #496, opened 2026-05-07
vorbis: Fix out of bounds panic during codebook codeword synthesis

baylesj:
Fuzzing discovered an index out-of-bounds panic when parsing heavily overspecified codebooks. The synthesize_codewords function attempted to evaluate next_codeword[i - 1] when i == 0, leading to integer underflow.

This updates the iteration range to 1..codeword_len + 1 and explicitly handles the i == 1 case to mirror the Vorbis reference logic without utilizing next_codeword[0], effectively averting the underflow condition.

IIUC, these codewords are part of the Huffman tree generation logic, so `next_codeword[0]` is effectively the root of the tree and thus has no parent and cannot be incremented, so its value does not need to be adjusted.

Tested and verified against malicious overspecified codebooks.

### DIFF
--- a/symphonia-codec-vorbis/src/codebook.rs
+++ b/symphonia-codec-vorbis/src/codebook.rs
@@ -146,14 +146,19 @@ fn synthesize_codewords(code_lens: &[u8]) -> Result<Vec<u32>> {
             return decode_error("vorbis: codebook overspecified");
         }
 
-        for i in (0..codeword_len + 1).rev() {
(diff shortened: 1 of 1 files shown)


baylesj:
@pdeljanov minor change fixing a corner case discovered through fuzzing.

pdeljanov:
@baylesj , thanks. No issues from testing on my end, and I see this brings us inline with the reference decoder. Merged!

PR #499, opened 2026-05-15
Parse channel count from opus atom

### LINKED ISSUE #498: Demuxing Opus from MP4 Doesn't Provide Channel Information from OpusAtom
When demuxing Opus packets from an MP4 container, the current OpusAtom doesn't parse any information out, like the channels, which are essential for decoding the packets.

tsar-boomba:
Implemented based on: [https://www.opus-codec.org/docs/opus_in_isobmff.html](https://www.opus-codec.org/docs/opus_in_isobmff.html)

Fixes #498

### DIFF
--- a/symphonia-format-isomp4/src/atoms/opus.rs
+++ b/symphonia-format-isomp4/src/atoms/opus.rs
@@ -71,6 +71,7 @@ impl Atom for OpusAtom {
 impl OpusAtom {
     pub fn fill_audio_sample_entry(self, entry: &mut AudioSampleEntry) -> Result<()> {
         entry.codec_id = CODEC_ID_OPUS;
+        entry.extra_data = Some(self.extra_data.clone());
 
         let mut reader = BufReader::new(&self.extra_data[OPUS_MAGIC_LEN..]);
 
--- a/symphonia-format-isomp4/src/atoms/opus.rs
+++ b/symphonia-format-isomp4/src/atoms/opus.rs
@@ -5,10 +5,12 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use symphonia_core::audio::{Channels, Position};
+use std::convert::identity;
+
+use symphonia_common::xiph::audio::opus;
 use symphonia_core::codecs::audio::well_known::CODEC_ID_OPUS;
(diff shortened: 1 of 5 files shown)


pdeljanov (changes requested):
Thanks for this. This is mostly correct, except we are going to need the extra data (i.e., the contents of this atom with the manually pre-pended "OpusHead" signature) for when we implement an Opus decoder. This is the identification header.

So you should the read function and extra data the same, and move your parsing code to into `fill_audio_sample_entry`. If you encounter an error, simply do noy fill the channels and sample rate.

symphonia-format-isomp4/src/atoms/opus.rs
```diff
@@ -66,8 +69,87 @@ impl Atom for OpusAtom {
 }
 
 impl OpusAtom {
-    pub fn fill_audio_sample_entry(self, entry: &mut AudioSampleEntry) {
+    pub fn fill_audio_sample_entry(self, entry: &mut AudioSampleEntry) -> Result<()> {
         entry.codec_id = CODEC_ID_OPUS;
-        entry.extra_data = Some(self.extra_data);
+        entry.extra_data = Some(self.extra_data.clone());
```
pdeljanov: No need to clone `self.extra_data` because we're consuming `self`. However, you will have to move it below the new parsing code.

symphonia-format-isomp4/src/atoms/stsd.rs
```diff
@@ -322,7 +322,7 @@ impl Atom for AudioSampleEntry {
                 }
                 AtomType::OpusDsConfig => {
                     let atom = it.read_atom::<OpusAtom>()?;
-                    atom.fill_audio_sample_entry(&mut entry);
+                    atom.fill_audio_sample_entry(&mut entry)?;
```
pdeljanov: This would cause malformed extra data for a track to make the entire file unplayable which is why all the other cases do not return an error. I think this one should not return an error either.

Take a look at `AlacAtom::fill_audio_sample_entry` in `alac.rs` for one way on how you could structure the solution.

pdeljanov (changes requested):
This is almost there, but `fill_audio_sample_entry` will need to ignore any extra data parsing errors or else this'll make the entire file unplayable if the extra data is broken.

pdeljanov:
Thanks, merged!

PR #500, opened 2026-05-16
Remove all unwraps

qarmin:
All unwrap() calls were converted into .expect() or proper error handling. 

I also added a Clippy rule to prevent accidentally introducing these patterns into the codebase.


symphonia-metadata/src/id3v2/frames/readers.rs
```diff
@@ -174,7 +177,10 @@ fn read_lang_code(reader: &mut BufReader<'_>) -> Result<Option<String>> {
     }
     else {
         // Convert to lowercase string.
-        Some(std::str::from_utf8(&code).unwrap().to_ascii_lowercase())
+        Some(match std::str::from_utf8(&code) {
```
pdeljanov: This can be changed to be an `except` because all characters are ASCII per-above check.

symphonia-metadata/src/id3v2/frames/readers.rs
```diff
@@ -145,7 +145,10 @@ fn read_date(reader: &mut BufReader<'_>) -> Result<String> {
     }
 
     // Safety: The data array only contains ASCII digits.
-    Ok(str::from_utf8(&date).unwrap().to_string())
+    match str::from_utf8(&date) {
```
pdeljanov: This can be changed to be an `except` because all characters are ASCII per-above check.

symphonia-format-riff/src/wave/mod.rs
```diff
@@ -103,11 +103,12 @@ impl<'s> WavReader<'s> {
 
             // The last chunk should always be a data chunk, if it is not, then the stream is
             // unsupported.
-            if chunk.is_none() {
-                return unsupported_error("wav: missing data chunk");
-            }
+            let chunk = match chunk {
+                Some(c) => c,
+                None => return unsupported_error("wav: missing data chunk"),
+            };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of matches.

symphonia-format-ogg/src/logical.rs
```diff
@@ -210,7 +210,11 @@ impl LogicalStream {
             let page_end_ts = self.mapper.absgp_to_ts(page.header.absgp);
 
             // Update previous page information with the timestamp.
-            self.prev_page_info.as_mut().unwrap().end_ts = Some(page_end_ts);
+            let prev_page = match self.prev_page_info.as_mut() {
+                Some(p) => p,
+                None => return decode_error("ogg: previous page info is missing"),
+            };
+            prev_page.end_ts = Some(page_end_ts);
```
pdeljanov: `self.prev_page_info` is unconditionally assigned in this function on line 133, so this should be an `expect()`.

symphonia-format-riff/src/aiff/mod.rs
```diff
@@ -136,7 +136,11 @@ impl<'s> AiffReader<'s> {
                     }
 
                     // The length of the sound data chunk must also be known.
-                    if let Some(len) = ssnd.as_ref().unwrap().len {
+                    let ssnd_ref = match ssnd.as_ref() {
```
pdeljanov: `ssnd` was assigned a few lines above at 133, use an `expect`.

symphonia-format-ogg/src/mappings/vorbis.rs
```diff
@@ -233,17 +233,21 @@ impl Mapper for VorbisMapper {
...
+                        Some(a) => a,
+                        None => {
+                            return decode_error("ogg (vorbis): codec parameters are not audio");
+                        }
+                    };
+                    let mut extra_data = match audio.extra_data.take() {
+                        Some(d) => d,
+                        None => return decode_error("ogg (vorbis): missing extra data"),
+                    }
+                    .into_vec();
```
pdeljanov: The safety comments  are correct. Mappers are created by the `detect` function in the mapper module which always populates this stuff. So these unwraps (and the ones below), will never fail. They can be made to be `expect()`, or, to keep things clean, use `.ok_ok(...)?` pattern.

symphonia-format-isomp4/src/atoms/esds.rs
```diff
@@ -50,11 +50,12 @@ impl Atom for EsdsAtom {
             }
         }
 
-        if descriptor.is_none() {
-            return decode_error("isomp4 (esds): missing es descriptor in esds");
-        }
+        let es_desc = match descriptor {
+            Some(d) => d,
+            None => return decode_error("isomp4 (esds): missing es descriptor in esds"),
+        };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of situations.

symphonia-common/src/mpeg/formats/mod.rs
```diff
@@ -213,16 +213,18 @@ impl ObjectDescriptor for ESDescriptor {
         scoped.ignore()?;
 
         // Decoder configuration descriptor is mandatory.
-        if dec_config.is_none() {
-            return decode_error("common (mp4): missing decoder config descriptor");
-        }
+        let dec_config = match dec_config {
+            Some(v) => v,
+            None => return decode_error("common (mp4): missing decoder config descriptor"),
+        };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of matches.

symphonia-format-isomp4/src/atoms/ilst.rs
```diff
@@ -319,8 +319,14 @@ fn add_pair_tag<R: ReadAtom>(
...
-            let d0 = u16::from_be_bytes(value.data[2..4].try_into().unwrap());
-            let d1 = u16::from_be_bytes(value.data[4..6].try_into().unwrap());
+            let d0 = u16::from_be_bytes(match value.data[2..4].try_into() {
+                Ok(b) => b,
+                Err(_) => return decode_error("isomp4 (ilst): track/disk slice is wrong size"),
+            });
+            let d1 = u16::from_be_bytes(match value.data[4..6].try_into() {
+                Ok(b) => b,
+                Err(_) => return decode_error("isomp4 (ilst): track/disk slice is wrong size"),
+            });
```
pdeljanov: These will clearly never fail. Use `expect`.

symphonia-format-isomp4/src/atoms/ilst.rs
```diff
@@ -470,7 +476,10 @@ fn add_id3v1_genre_tag<R: ReadAtom>(
         // The ID3v1 genre number is stored as a unsigned 16-bit big-endian integer. The data type
         // is no-type.
         if value_atom.data.len() == 2 {
-            let index = u16::from_be_bytes(value_atom.data.as_ref().try_into().unwrap());
+            let index = u16::from_be_bytes(match value_atom.data.as_ref().try_into() {
+                Ok(b) => b,
+                Err(_) => return decode_error("isomp4 (ilst): genre data is wrong size"),
+            });
```
pdeljanov: This will clearly never fail, use `expect`.

symphonia-format-isomp4/src/atoms/mdia.rs
```diff
@@ -39,18 +39,19 @@ impl Atom for MdiaAtom {
...
+                None => return decode_error("isomp4 (mdia): missing mdhd atom"),
+            },
+            hdlr: match hdlr {
+                Some(v) => v,
+                None => return decode_error("isomp4 (mdia): missing hdlr atom"),
+            },
+            minf: match minf {
+                Some(v) => v,
+                None => return decode_error("isomp4 (mdia): missing minf atom"),
+            },
```
pdeljanov: Use the `.ok_or(...)?` pattern, for these types of matches.

symphonia-format-isomp4/src/atoms/minf.rs
```diff
@@ -36,10 +36,12 @@ impl Atom for MinfAtom {
...
-            return decode_error("isomp4 (minf): missing stbl atom");
-        }
-
-        Ok(MinfAtom { smhd, stbl: stbl.unwrap() })
+        Ok(MinfAtom {
+            smhd,
+            stbl: match stbl {
+                Some(v) => v,
+                None => return decode_error("isomp4 (minf): missing stbl atom"),
+            },
```
pdeljanov: Use the `.ok_or(...)?` pattern, for these types of matches.

symphonia-format-isomp4/src/atoms/moof.rs
```diff
@@ -40,13 +40,14 @@ impl Atom for MoofAtom {
             }
         }
 
-        if mfhd.is_none() {
-            return decode_error("isomp4 (moof): missing mfhd atom");
-        }
+        let mfhd = match mfhd {
+            Some(v) => v,
+            None => return decode_error("isomp4 (moof): missing mfhd atom"),
+        };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of situations.

symphonia-format-isomp4/src/atoms/moov.rs
```diff
@@ -66,9 +66,10 @@ impl Atom for MoovAtom {
             }
         }
 
-        if mvhd.is_none() {
-            return decode_error("isomp4 (moov): missing mvhd atom");
-        }
+        let mvhd = match mvhd {
+            Some(v) => v,
+            None => return decode_error("isomp4 (moov): missing mvhd atom"),
+        };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of situations.

symphonia-format-isomp4/src/atoms/stbl.rs
```diff
@@ -65,32 +65,28 @@ impl Atom for StblAtom {
...
+                None => return decode_error("isomp4 (stbl): missing stts atom"),
+            },
+            stsc: match stsc {
+                Some(v) => v,
+                None => return decode_error("isomp4 (stbl): missing stsc atom"),
+            },
+            stsz: match stsz {
+                Some(v) => v,
+                None => return decode_error("isomp4 (stbl): missing stsz atom"),
+            },
```
pdeljanov: Use the `.ok_or(...)?` pattern for these types of situations.

symphonia-format-isomp4/src/atoms/traf.rs
```diff
@@ -46,10 +46,11 @@ impl Atom for TrafAtom {
         }
 
         // Tfhd is mandatory.
-        if tfhd.is_none() {
-            return decode_error("isomp4 (traf): missing tfhd atom");
-        }
+        let tfhd = match tfhd {
+            Some(v) => v,
+            None => return decode_error("isomp4 (traf): missing tfhd atom"),
+        };
```
pdeljanov: Use a `let-else` expression, or replace `.unwrap()` with `.ok_or(...)?`, for these types of situations.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -273,7 +273,10 @@ impl<'s> IsoMp4Reader<'s> {
             }
         }
 
-        let mut moov = moov.unwrap();
+        let mut moov = match moov {
```
pdeljanov: We bail out if `moov` is `None` above, so this will never fail. Can be made `expect`.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -351,7 +354,10 @@ impl<'s> IsoMp4Reader<'s> {
 
         for (state, track) in self.track_states.iter().zip(&self.tracks) {
             // Get the timebase of the track used to calculate the presentation time.
-            let tb = track.time_base.unwrap();
+            let tb = match track.time_base {
+                Some(tb) => tb,
+                None => return decode_error("isomp4: track is missing time base"),
+            };
```
pdeljanov: A track is always created with a timebase, can be `expect`.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -437,7 +443,11 @@ impl<'s> IsoMp4Reader<'s> {
         //
         // Note, there will always be one segment because the moov atom was converted into a segment
         // when the reader was instantiated.
-        if self.segs.last().unwrap().all_tracks_ended() {
+        let last_seg = match self.segs.last() {
+            Some(s) => s,
+            None => return decode_error("isomp4: no segments available"),
+        };
```
pdeljanov: Can be made `expect`, there is always one segment from the `moov` atom.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -468,7 +478,10 @@ impl<'s> IsoMp4Reader<'s> {
                     // A moof segment can only be created if the media is fragmented.
                     if self.moov.is_fragmented() {
                         // Get the last segment.
-                        let last_seg = self.segs.last().unwrap();
+                        let last_seg = match self.segs.last() {
```
pdeljanov: Can be made `expect`, there is always one segment from the `moov` atom.

symphonia-format-isomp4/src/demuxer.rs
```diff
@@ -498,7 +511,10 @@ impl<'s> IsoMp4Reader<'s> {
     fn seek_track_by_time(&mut self, track_num: usize, time: Time) -> Result<SeekedTo> {
         // Convert time to timestamp for the track.
         if let Some(track) = self.tracks.get(track_num) {
-            let tb = track.time_base.unwrap();
+            let tb = match track.time_base {
```
pdeljanov: A track always has a timebase, can be made `expect`.

symphonia-core/src/io/bit.rs
```diff
@@ -266,7 +266,15 @@ pub mod vlc {
                 let mut entry_count = 0;
 
                 // Get the block id at the front of the queue.
-                let block_id = queue.pop_front().unwrap();
+                let block_id = match queue.pop_front() {
```
pdeljanov: The loop condition is that `!queue.is_empty()` so this will never fail. However, the proper solution is to change the while loop to `while let Some(block_id) = queue.pop_front()`.

pdeljanov (changes requested):
Thanks looking into removing unwrap, I think it is a worthwhile thing to do. At a minimum, `expect` should be the baseline.

I left a lot of comments here because there are two categories of issues:

1. Unwrap was used in a place where it is very obvious it will never fail. In these cases I say we should use `expect`.
2. Excessive use of `match` where a `let-else` expression or `.ok_ok(...)?` pattern would keep the code more compact and tidier.

There are probably more cases where an `expect` would be fine, however, if it wasn't immediately obvious, I was okay with leaving returning an error.

symphonia-format-ogg/src/logical.rs
```diff
@@ -527,7 +534,10 @@ impl LogicalStream {
         let page_end_ts = self.mapper.absgp_to_ts(page.header.absgp);
 
         // Map the absolute granule position to a timestamp.
-        let page_start_ts_raw = page_end_ts.checked_sub(total_pkt_dur).unwrap();
+        let page_start_ts_raw = match page_end_ts.checked_sub(total_pkt_dur) {
+            Some(ts) => ts,
+            None => return decode_error("ogg: timestamp underflow"),
+        };
```
pdeljanov: I think this should be changed to a `saturating_sub` and this function should not return an error so it doesn't break the binary-search in `do_seek`.

symphonia-format-riff/src/aiff/mod.rs
```diff
@@ -265,8 +266,11 @@ fn process_markers(
             }
 
             // Add the chapter.
+            let start_time = tb
+                .calc_time(Timestamp::from(marker.ts))
+                .ok_or(Error::DecodeError("aiff: chapter timestamp overflow"))?;
```
pdeljanov: No possibility for overflow because `marker.ts` is `u32`. The computation being performed is `marker.ts / comm.sample_rate` to calculate the number of seconds. `Time` represents seconds as an `i64`, so there is no chance for an overflow. Therefore, this can be an `expect` and the function

pdeljanov (changes requested):
Thanks. Found a couple small issues. Aside from that, LGTM.

I see that the CI is failing on Mac/Windows because there are a few source files that are not built on Linux that contain `unwrap`. Do you need help fixing those?

pdeljanov (approved):

pdeljanov:
Thanks, merged!

Issue #505, opened 2026-05-25
const lookup tables are causing unneccessary memcpys in debug builds


bolshoytoster:
I've been profiling my audio thread, and I've noticed it spends ~50% of its time doing memcpys in `symphonia-bundle-mp3`'s `synthesis` function, in these 2 lines:
https://github.com/pdeljanov/Symphonia/blob/8abf8fe7c5a11e772883a7510d65f47eb5147c11/symphonia-bundle-mp3/src/synthesis.rs#L320-L321

This is due to it having to copy `SYNTHESIS_D` every time it's indexed. There's more info in [this rust issue](https://github.com/rust-lang/rust/issues/156911)

The solution would be to either take the indexes by reference (i.e. `&SYNTHESIS_D[k + i + 0];`), change the lookup table const to a `&[f32; 512]`/`&[f32]`, or change it to a static.

There seem to be quite a few other lookup tables in the repo that might benefit from this (I've found 149 potential from a [quick regex search](https://github.com/search?q=repo%3Apdeljanov%2FSymphonia+%2Fconst+%5BA-Z_0-9%5D*%3A+%5C%5B%2F)).

I can open a PR to fix these if you want.

pdeljanov:
Well this is surprising. Thanks for reporting.

Based on the Rust [reference](https://doc.rust-lang.org/reference/items/static-items.html#r-items.static.alternate), `const` should be preferred unless the constant is "large". Which is a bit open-ended. What is "large"? Would there be a negative consequence for making all tables `static`? I suppose if the table elements are indexed by a constant, or a value computed at compile-time, `const` would be preferred so that it may be inlined, at least for release builds.

Thanks for creating the PR. I'll experiment a bit.

bolshoytoster:
> Would there be a negative consequence for making all tables static? I suppose if the table elements are indexed by a constant, or a value computed at compile-time, const would be preferred so that it may be inlined, at least for release builds.

[...]

PR #507, opened 2026-05-26
Change some of the biggest lookup tables in the repo from consts to statics

bolshoytoster:
See #505 for details. I chose to update only a few tables that are >1024 bytes, just for the sake of creating a small diff. I skipped a few larger tables that wouldn't benefit from this.

There are definitely more tables that would benefit from this - and I could update more if you want - but the smaller ones don't matter as much.


baylesj:
This change **LGTM**, but I don't have approval rights. We should also consider updating the [clippy.toml](https://github.com/pdeljanov/Symphonia/blob/main/clippy.toml) file to have a stricter limit. My vote would be `256`:

```toml
array-size-threshold = 256
```

@pdeljanov

pdeljanov:
@baylesj I added that Clippy lint and the warnings looked very reasonable. I then made the recommended changes and did some benchmarks and saw no meaningful performance difference in release mode. The release binary shaved off an additional 16 kB.

@bolshoytoster Let me know if you are able to update the lint and apply the recommended fixes. Otherwise, I can merge this as-is and do a follow-up commit.

bolshoytoster:
I've updated the lint.

pdeljanov:
Thanks!

PR #511, opened 2026-06-05
mp3 (layer3): accept main_data offset at buffer end for zero-length

opera-emoller:
It's valid for MP3s to contain zero-byte-length channels and they can end up last in the buffer.

### DIFF
--- a/symphonia-bundle-mp3/src/layer3/mod.rs
+++ b/symphonia-bundle-mp3/src/layer3/mod.rs
@@ -305,7 +305,8 @@ impl Layer3 {
                 let byte_index = part2_3_begin >> 3;
 
                 // Create a bit reader at the expected starting bit position.
-                let mut bs = if byte_index < main_data.len() {
+                // A zero byte silent channel can begin exactly at main_data.len().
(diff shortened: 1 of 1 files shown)


symphonia-bundle-mp3/src/layer3/mod.rs
```diff
@@ -305,7 +305,8 @@ impl Layer3 {
                 let byte_index = part2_3_begin >> 3;
 
                 // Create a bit reader at the expected starting bit position.
-                let mut bs = if byte_index < main_data.len() {
+                // A zero byte silent channel can begin exactly at main_data.len().
+                let mut bs = if byte_index <= main_data.len() {
```
baylesj, line 309: This condition is correct as far as I can tell in the ISO spec, but I would suggest adding either a sample file or an in-module unit test for regression test coverage.

baylesj (commented):
*LGTM* % having either a sample file or an in-module test would be helpful to ensure there are no regressions here.

pdeljanov:
It looks like this got duplicated in #528. However, I'm happy both PRs landed on the same solution. Thank you for opening this!

PR #513, opened 2026-06-10
Fix out-of-bounds index panic in AAC section data decoding

### LINKED ISSUE #512: Index-out-of-bounds in symphonia-codec-aac 0.6.0 at aac/ics/mod.rs:246 (decode_section_data) — distinct from #424
### Description

Decoding a crafted AAC (ADTS) stream panics with an index-out-of-bounds in
`symphonia-codec-aac` **0.6.0** (latest crates.io release), inside
`Ics::decode_section_data`.

This is a **distinct, still-present sibling** of the AAC spectrum bug #424
(which was at `ics/mod.rs:365` in `decode_spectrum` and is fixed in 0.6) and of
the ADTS bug #509 (`adts.rs:303`, fixed in 0.6). This one is in the
*section-data* parser, not spectrum decoding, and reproduces on 0.6.0.

### Crash location

```
symphonia-codec-aac-0.6.0/src/aac/ics/mod.rs:246:17  (Ics::decode_section_data)
```

### Error message

```
thread 'main' panicked at .../symphonia-codec-aac-0.6.0/src/aac/ics/mod.rs:246:17:
index out of bounds: the len is 64 but the index is 64
```

Backtrace (trimmed):

```
ics/mod.rs:246  Ics::decode_section_data
ics/mod.rs (decode)  -> cpe.rs:58  decode_ga_sce
aac/mod.rs:137  decode_ga
aac/mod.rs:240, :265
codecs/audio.rs:280  <AacDecoder as AudioDecoder>::decode
```

### Likely root cause

`sect_cb`/`sect_len` are fixed-size `[[_; MAX_SFBS]; MAX_WINDOWS]` with
`MAX_SFBS == 64` (mod.rs:203-204). In `decode_section_data`:

```rust
while k < self.info.max_sfb {
    self.sect_cb[g][l] = bs.read_bits_leq32(4)? as u8;   // <-- line 246, l can reach 64
    self.sect_len[g][l] = 0;
    ...
    // l += 1 once per section; k advances by the section length
}
```

`l` is the *section* index and is incremented once per section, but it is never
bounded by `MAX_SFBS`. A crafted stream that produces more sections than bands
(e.g. many zero-length sections) drives `l` up to `64`, indexing one past the
end of `sect_cb[g]`. `max_sfb` itself is read as up to 6 bits (mod.rs:166), and
the loop terminates on the *band* count `k`, not on `l`, so validating `max_sfb`
alone does not bound `l`.

### Environment

- `symphonia-codec-aac` 0.6.0 (via `symphonia` 0.6.0)
- rustc 1.95.0, aarch64-apple-darwin
- Reproduces in both debug and release (it is a slice bounds check, not an
  overflow check), and without any sanitizer.

### Self-contained reproducer

`Cargo.toml`:

```toml
[package]
name = "sym-repro"
version = "0.0.0"
edition = "2021"

[dependencies]
symphonia = { version = "0.6.0", default-features = false, features = ["aac", "isomp4"] }
```

`src/main.rs` — mirrors a streaming player (probe, pull interleaved f32 in
fixed chunks, seek, pull again); the crafted input is inlined:

```rust
use std::io::Cursor;
use symphonia::core::codecs::audio::{AudioDecoder, AudioDecoderOptions};
use symphonia::core::formats::probe::Hint;
use symphonia::core::formats::{FormatOptions, FormatReader, SeekMode, SeekTo, TrackType};
use symphonia::core::io::MediaSourceStream;
use symphonia::core::meta::MetadataOptions;
use symphonia::core::units::Timestamp;

const DATA: &[u8] = &[
    0xff, 0xf3, 0xff, 0xff, 0xff, 0xf8, 0x00, 0x80, 0x01, 0xff, 0xf8, 0xff, 0xf8, 0xff, 0xef, 0x01,
    0xe9, 0x00, 0x00, 0xff, 0xf8, 0x00, 0xc4, 0x12, 0x01, 0x80, 0x3d, 0xff, 0x10, 0x00, 0x04, 0x00,
    0x00, 0x00, 0x05, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff, 0x56, 0x00, 0x00, 0x00, 0x01,
    0x20, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x2b, 0x2b, 0x2b, 0x00, 0x00, 0x00, 0x00,
    0x00, 0x00, 0x00, 0x01, 0xff, 0xf8, 0xff, 0xf8, 0xf7, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x5d,
    0x00, 0x00, 0x00, 0x01, 0x7f, 0x7f, 0xff, 0x30, 0xf8, 0x00, 0x80, 0x01, 0xff, 0xf8, 0xff, 0xf8,
    0x28, 0xef, 0xff, 0xcf,
];

struct Dec {
    fmt: Box<dyn FormatReader>,
    dec: Box<dyn AudioDecoder>,
    track_id: u32,
    leftover: Vec<f32>,
    pos: usize,
}

impl Dec {
    fn open(bytes: Vec<u8>) -> Option<Self> {
        let mss = MediaSourceStream::new(Box::new(Cursor::new(bytes)), Default::default());
        let fmt = symphonia::default::get_probe()
            .probe(&Hint::new(), mss, FormatOptions::default(), MetadataOptions::default())
            .ok()?;
        let track = fmt.default_track(TrackType::Audio)?.clone();
        let track_id = track.id;
        let params = track.codec_params.as_ref()?.audio()?.clone();
        let dec = symphonia::default::get_codecs()
            .make_audio_decoder(&params, &AudioDecoderOptions::default())
            .ok()?;
        Some(Self { fmt, dec, track_id, leftover: Vec::new(), pos: 0 })
    }

    fn read(&mut self, out: &mut [f32]) -> usize {
        let mut w = 0;
        while w < out.len() {
            if self.pos < self.leftover.len() {
                let n = (out.len() - w).min(self.leftover.len() - self.pos);
                out[w..w + n].copy_from_slice(&self.leftover[self.pos..self.pos + n]);
                self.pos += n;
                w += n;
                continue;
            }
            self.leftover.clear();
            self.pos = 0;
            let packet = loop {
                match self.fmt.next_packet() {
                    Ok(Some(p)) if p.track_id == self.track_id => break p,
                    Ok(Some(_)) => continue,
                    _ => return w,
                }
            };
            match self.dec.decode(&packet) {
                Ok(b) => b.copy_to_vec_interleaved::<f32>(&mut self.leftover),
                Err(symphonia::core::errors::Error::DecodeError(_)) => continue,
                Err(_) => return w,
            }
        }
        w
    }

    fn seek(&mut self, frame: u64) {
        let to = SeekTo::Timestamp { ts: Timestamp::new(frame as i64), track_id: self.track_id };
        if self.fmt.seek(SeekMode::Accurate, to).is_ok() {
            self.dec.reset();
            self.leftover.clear();
            self.pos = 0;
        }
    }
}

fn main() {
    let Some(mut d) = Dec::open(DATA.to_vec()) else { println!("unsupported"); return; };
    let mut buf = vec![0f32; 4096];
    for _ in 0..256 {
        if d.read(&mut buf) == 0 { break; }
    }
    d.seek(1000);
    d.read(&mut buf); // panics: index out of bounds: the len is 64 but the index is 64
    println!("no panic (unexpected)");
}
```

`cargo run` (or `cargo run --release`) panics.

Found by fuzzing the decoder with `cargo-fuzz`. Thanks for Symphonia!

baylesj:
Resolves an issue where a crafted stream can produce many zero-length sections, causing the section index `l` to exceed `MAX_SFBS` and panic when accessing `sect_cb` and `sect_len`.

Fixes #512.

### DIFF
--- a/symphonia-codec-aac/src/aac/ics/mod.rs
+++ b/symphonia-codec-aac/src/aac/ics/mod.rs
@@ -240,6 +240,8 @@ impl Ics {
             let mut l = 0;
 
             while k < self.info.max_sfb {
+                validate!(l < MAX_SFBS);
+
                 self.sect_cb[g][l] = bs.read_bits_leq32(4)? as u8;
                 self.sect_len[g][l] = 0;
 


baylesj:
@pdeljanov easy enough fix for the OOB, PTAL.

pdeljanov:
Seems like we had a duplicate fix for this in #537. It had a very descriptive comment and test so I just opted to merge that one instead. Thank you though!

PR #515, opened 2026-06-10
riff: gracefully handle malformed child chunk lengths

### LINKED ISSUE #380: wav file that works elsewhere results in "riff: chunk length exceeds parent (list) chunk length"
Symphonia fails to read the following wav file during
```rs
get_probe().format(&Hint::new(), mss, &Default::default(), &Default::default())
```

the file plays without problems at least in mac's finder, vlc and audacity.

https://www.dropbox.com/scl/fi/xjwgq0euy0uleuaaqdwrx/koala.wav?rlkey=xz8fry878e31vqdne6s0nkpbp&st=em18o14d&dl=0

baylesj:
Instead of returning a hard decode error when a chunk length exceeds the remaining space in the parent list chunk, we now log a warning and truncate the chunk length to fit within the parent. This allows malformed WAV files that have incorrect RIFF lengths to still be played correctly.

Fixes #380

### DIFF
--- a/symphonia-format-riff/src/common.rs
+++ b/symphonia-format-riff/src/common.rs
@@ -20,7 +20,7 @@ use symphonia_core::errors::{Result, decode_error};
 use symphonia_core::formats::prelude::*;
 use symphonia_core::io::{MediaSourceStream, ReadBytes};
 
-use log::{debug, info};
+use log::{info, warn};
 
 pub enum ByteOrder {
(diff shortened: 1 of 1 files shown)


baylesj:
@pdeljanov  this seems to me to be a reasonable compromise between being correct in behavior and allowing for the playing of files that aren't rigidly correct. LMK what you think.

Issue #516, opened 2026-06-11
mp3: Xing/Info tag rejected on CRC-protected frames, VBR duration estimate ends up way off


f-o-o-g-s:
While chasing a duration bug in my player (a 4:44 VBR mp3 displaying as 30:24) I traced it down to `is_maybe_info_tag` in symphonia-bundle-mp3. The fast check wants every byte between the frame header and the tag ID to be zero:

```rust
// The side information should be zeroed.
!buf[MPEG_HEADER_LEN..offset].iter().any(|&b| b != 0)
```

But when the frame has the protection bit set, the 16 bit frame CRC sits right after the 4 byte header, inside that scanned range, and it's basically never zero. LAME writes the Xing ID at side_info_len() + 4 regardless of CRC, the same offset symphonia looks at, so the tag is otherwise perfectly parseable. The check fails, the whole Xing tag gets thrown away, and the demuxer falls back to estimate_num_mpeg_frames. That averages the first 17 frames and extrapolates over the file, so a VBR file that starts with silence (32kbps frames) gets a duration several times too long. My real file is 4:44 at ~240kbps average (LAME 3.97, valid Xing tag with frames=10885, valid LAME tag CRC) and symphonia reports it as 1824.757s.

Repro from scratch, needs sox and lame:

```
sox -D -n -r 44100 -c 2 -b 16 quiet.wav trim 0 3
sox -D -n -r 44100 -c 2 -b 16 loud.wav synth 27 noise vol 0.7
sox -D quiet.wav loud.wav full.wav
lame -p -V 2 full.wav repro.mp3
```

lame's `-p` is what turns on the frame CRCs. The result is a 30 second file that probes at 80 to 95 seconds (exact number varies with the random noise). I get:

```
0.5.5:   n_frames Some(4161024) -> 94.354s
0.6.0:   94.354s (the check is unchanged on main as well)
ffprobe: 30.0s, and the Xing tag itself says frames=1150 which is right
```

I tried a one line fix on a checkout of v0.6.0, skipping the CRC bytes in the scan:

```rust
let start = MPEG_HEADER_LEN + if header.has_crc { 2 } else { 0 };

!buf[start..offset].iter().any(|&b| b != 0)
```

With that the repro file probes 30.000s and my real file 284.280s, both matching ffmpeg. The tag CRC validation inside try_read_info_tag_inner already hashes from the start of the frame including the frame CRC bytes, so a valid LAME tag CRC still checks out after this change (my real file's does).

[...]

PR #517, opened 2026-06-13
mp3: Skip CRC in Xing/Info and VBRI tag heuristics

baylesj:
This PR addresses issue #516 - MP3 Xing/Info tags were rejected on CRC-protected frames. 

The heuristic checks in `is_maybe_info_tag` and `is_maybe_vbri_tag` were scanning the side information for zeros starting immediately after the 4-byte MPEG header. However, if the frame has the protection bit set, a 2-byte CRC follows the header. This CRC is almost never zero when utilized, causing the check to fail and the tag to be discarded. Although most MP3 files do not use CRC protection, this is a tightly scoped fix that makes Symphonia more compliant by checking the scan range to start after the CRC if it is present.

Verified with new in-module unit tests in the mp3 `demuxer.rs`. I would prefer a sample file to avoid regressions but do not have one readily available.

### DIFF
--- a/symphonia-bundle-mp3/src/demuxer.rs
+++ b/symphonia-bundle-mp3/src/demuxer.rs
@@ -1080,8 +1080,7 @@ mod tests {
         // Tag ID at offset 32 + 4 = 36
         buf[36..40].copy_from_slice(b"Xing");
 
-        // Currently this should return false because of the CRC, but it should return true.
-        // We want this to return true.
+        // Ensure the heuristic returns true even if a non-zero CRC is present.
         assert!(is_maybe_info_tag(&buf, &header));
     }
 
@@ -1111,7 +1110,7 @@ mod tests {
         // Tag ID at offset 36
         buf[36..40].copy_from_slice(b"VBRI");
 
-        // Currently this should return false because of the CRC, but it should return true.
+        // Ensure the heuristic returns true even if a non-zero CRC is present.
         assert!(is_maybe_vbri_tag(&buf, &header));
     }
 }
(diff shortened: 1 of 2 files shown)


baylesj:
@pdeljanov it would be nice to have a corpus or set of test files that we can use for validation somewhere (maybe hosted in a separate GitHub with LFS, on some kind of cloud storage, or something else), maybe simliar to how benchtest.py downloads things from `archive.org`. I'll give it some more thought. I think my interest would be in avoiding regressions & improving confidence in doing major refactors.

I've created an issue to have a place to discuss this:

https://github.com/pdeljanov/Symphonia/issues/518

PTAL

symphonia-bundle-mp3/src/demuxer.rs
```diff
@@ -964,7 +964,9 @@ fn is_maybe_info_tag(buf: &[u8], header: &FrameHeader) -> bool {
     }
 
     // The side information should be zeroed.
-    !buf[MPEG_HEADER_LEN..offset].iter().any(|&b| b != 0)
+    let start = MPEG_HEADER_LEN + if header.has_crc { 2 } else { 0 };
+
+    !buf[start..offset].iter().any(|&b| b != 0)
```
pdeljanov: This is a nit, but I'd recommend adding a tiny helper function to `FrameHeader`:

```rust
/// The size of the header itself including the optional CRC.
#[inline(always)]
pub fn header_size(&self) -> usize {
    MPEG_HEADER_LEN + if self.has_crc { 2 } else { 0 }
}
```

Then just use `header.header_size()` directly instead of start when slicing the buffer for both changed functions.

PR looks good otherwise!

pdeljanov (changes requested):

symphonia-bundle-mp3/src/demuxer.rs
baylesj: Works for me.

pdeljanov:
Thanks!

PR #520, opened 2026-06-13
core: clamp scan_bytes_aligned_ref to scan_len to prevent overrun and panic

### LINKED ISSUE #519: BufReader::scan_bytes_aligned overruns scan_len and can panic
## Description

`BufReader::scan_bytes_aligned_ref` (`symphonia-core/src/io/buf_reader.rs`) is documented to scan up to `scan_len` bytes. On the no-match path the window end advances by `align` per step and can pass the `scan_len` boundary by up to `align - 1` bytes, and the final position is clamped to the buffer length rather than that boundary:

```rust
self.pos = cmp::min(j, self.buf.len());
```

So when `align > 1`, `scan_len < remaining`, and the pattern is not found, the scan consumes and returns more than `scan_len` bytes. Through the public `ReadBytes::scan_bytes_aligned(pattern, align, buf)`, which passes `buf.len()` as `scan_len` and then copies the returned slice into `buf`, the over-long slice triggers a panic. CONTRIBUTING states Symphonia parses untrusted data and must never panic.

## Steps to Reproduce

```rust
use symphonia_core::io::{BufReader, ReadBytes};

// returns more than scan_len
let mut r = BufReader::new(b"ABCDEF");
let s = r.scan_bytes_aligned_ref(&[0x00, 0x00], 2, 3).unwrap();
assert_eq!(s.len(), 4); // scan_len was 3

// panics
let mut r = BufReader::new(b"ABCDEF");
let mut dst = [0u8; 3];
let _ = r.scan_bytes_aligned(&[0x00, 0x00], 2, &mut dst).unwrap();
// thread panicked: range end index 4 out of range for slice of length 3
```

## Expected Behavior

The scan returns at most `scan_len` bytes (`b"ABC"`, pos 3) and `scan_bytes_aligned` does not panic.

## Fix

Clamp the post-scan position to the scan boundary instead of the buffer length. In-tree callers pass `scan_len` equal to the remaining length, so they are unaffected.

## Environment

- crate: `symphonia-core`
- branch: `main`

Noethix55555:
## Summary

`scan_bytes_aligned_ref` could advance past the `scan_len` boundary on the no-match path and clamped the position to the buffer length, so it returned more than `scan_len` bytes. Via `ReadBytes::scan_bytes_aligned` this panicked when the destination buffer was smaller than the remaining input. Clamp to the scan boundary instead.

Existing in-tree callers pass `scan_len` equal to the remaining length, so they are unaffected. Adds regression tests to buf_reader.rs (which had none) covering pattern-found, no-match remainder, scan_len respected, and the small-buffer no-panic case.

Closes #519.

### DIFF
--- a/symphonia-core/src/io/buf_reader.rs
+++ b/symphonia-core/src/io/buf_reader.rs
@@ -72,7 +72,9 @@ impl<'a> BufReader<'a> {
             i += align;
         }
 
-        self.pos = cmp::min(j, self.buf.len());
+        // When the pattern is not matched, `j` may have advanced past `end` by up to `align - 1`
+        // bytes. Clamp to `end` so the scan never consumes more than `scan_len` bytes.
+        self.pos = cmp::min(j, end);
         Ok(&self.buf[start..self.pos])
     }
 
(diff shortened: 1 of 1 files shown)


baylesj:
*LGTM*, this is a tightly scoped fix that makes this class behave more in line with its API. I agree that the existing callers are unaffected, and the test coverage is a good add in my opinion.

@pdeljanov

Bolt4243:
thanks

pdeljanov:
Thank you @Noethix55555 , also LGTM, merged.

PR #528, opened 2026-07-09
layer3: Fix decoding failure on zero-length channels

### LINKED ISSUE #529: Symphonia fails to decode some MP3 files with non-standard silent audio frames
Issue discovered downstream in Chromium, see https://crbug.com/532860173.

In short, some encoders, such as the one included in ProTools, can generate audio files with non-standard silent audio frames. Symphonia fails to decode these frames properly / as expected, cropping out the silent frames entirely. Re-encoding the files with FFmpeg makes them play back normally with Symphonia, but this is not ideal.

[silent-section.mp3](https://github.com/user-attachments/files/29865711/silent-section.mp3)

baylesj:
If a channel has part2_3_length == 0, skip creating the BitReaderLtr. This avoids an out-of-bounds check (mpa: invalid main_data offset) when the bit reservoir is exactly exhausted at granule boundaries.

This resolves decoding some non-standard MP3 files. For more context, see the related issue in Chromium, [cl/532860173](https://crbug.com/532860173).

Fixes #529

### DIFF
--- a/symphonia-bundle-mp3/src/layer3/mod.rs
+++ b/symphonia-bundle-mp3/src/layer3/mod.rs
@@ -305,7 +305,7 @@ impl Layer3 {
                 let byte_index = part2_3_begin >> 3;
 
                 // Create a bit reader at the expected starting bit position.
-                let mut bs = if byte_index < main_data.len() {
+                let mut bs = if byte_index <= main_data.len() {
                     let mut bs = BitReaderLtr::new(&main_data[byte_index..]);
 
                     let bit_index = part2_3_begin & 0x7;


baylesj:
I think this is subject to the same clippy errors as my last patch -- I will attempt to resolve before tagging this for review.

pdeljanov:
This is reasonable and very tolerant solution. I'm open to merging it.

However, and please check my logic, isn't the solution to change `byte_index < main_data.len()` to `byte_index <= main_data.len()`? This resolves the reported issue, but is overall less tolerant, in that, if the encoder set `part2_3_length` to `0` but encodes scale factors (`part2_len` > `0`), it would return a decode error rather than silence.

Probably the more tolerant solution is better here? On the other hand, for that error case, it's ambiguous what the encoder wanted to do.

pdeljanov:
Ah, to be clear, I was proposing switching `<` to `<=` as the only required change if the logic checked out with you. I believe this change would prevent returning a decode error in a properly encoded file with `part2_3_length` of 0 while also not special-casing it.

baylesj:
Ahh, yeah removing the special case logic entirely should work fine and makes the code easier to maintain. Thanks for clarifying && I updated the PR to match.

pdeljanov:
Thanks!

PR #533, opened 2026-07-11
flac: recover the final frame past an appended ID3v1/APEv2 tag

attilagyorffy:
## Summary

Fixes FLAC files that fail to demux because a metadata tag (an ID3v1 or APEv2 tag) has been wrongly appended after the audio.

FLAC frames carry no length, so the packet parser delimits a frame by scanning for the *next* frame's sync word. When a trailing tag sits where that sync word would be, the tag gets merged onto the last fragment; its CRC-16 footer no longer matches, the final frame is never built, and — because `STREAMINFO` declares more samples than were demuxed — `next_packet` returns `IoError(UnexpectedEof)` rather than the final packet. The whole track fails to demux (probe succeeds, first `next_packet` after the penultimate frame errors).

Such files are common in the wild: taggers that also write ID3 to MP3 sometimes prepend an ID3v2 tag and append an ID3v1 tag to FLAC too.

## What this does

On end-of-stream, before treating a sample shortfall as truncation, attempt to recover the final frame:

1. Reassemble the pending fragments into one buffer.
2. Strip a recognised trailing tag — a 128-byte ID3v1 tag (`"TAG"` magic), or an APEv2 tag located from its 32-byte `"APETAGEX"` footer.
3. Accept the recovery **only if** the trimmed buffer's trailing two bytes are the CRC-16 of everything before them.

That CRC check is the safety net: it distinguishes a genuine trailing tag (the frame's own CRC-16 is now exposed and matches) from a truncated final frame (no match → the existing `UnexpectedEof` truncation error still fires). Clean streams are unaffected — their final frame is already delimited by end-of-stream.

## Testing

- **Unit tests** (self-contained, no fixtures): trailing-tag detection for ID3v1 and APEv2, an untagged buffer, and the frame recovery behind an ID3v1 tag (using a synthetic frame + valid CRC-16), plus the negative case (a clean frame is not "recovered").
- **Real files**: verified against ID3v2-prefixed / ID3v1-suffixed FLAC files (19–52 MB) that error on current `main`. After the fix they demux to completion — e.g. one file goes from 1749 packets + `UnexpectedEof` to a clean 1750 packets, matching ffmpeg's frame count — and the recovered final frame is byte-identical to the frame a tag-free copy yields.

## Notes (draft)

- Recognises ID3v1 and APEv2 trailers; other trailing junk still errors (safely), which could be extended later.
- Opened as a draft for feedback on the approach.

### DIFF
--- a/symphonia-bundle-flac/src/parser.rs
+++ b/symphonia-bundle-flac/src/parser.rs
@@ -276,10 +276,75 @@ impl PacketBuilder {
         Some(ParsedPacket { buf: data, sync })
     }
 
+    /// At end-of-stream, attempt to recover a final frame whose delimiter — the next frame's sync
+    /// word — was replaced by a metadata tag wrongly appended after the audio (e.g. an ID3v1 or
+    /// APEv2 tag). Such a tag gets merged onto the last fragment, breaking its CRC-16 footer so the
+    /// frame is never built. Removing the tag re-exposes the frame's own CRC-16, letting it build.
+    /// Returns `None` when there is nothing to recover or the trimmed data is not a valid frame, in
+    /// which case the caller falls back to its normal truncation handling.
+    fn recover_final_fragment(&mut self) -> Option<Fragment> {
+        if self.frags.is_empty() {
+            return None;
+        }
+
+        // Reassemble the pending fragments into a single buffer.
+        let mut buf = Vec::new();
+        for frag in &self.frags {
+            buf.extend_from_slice(&frag.data);
+        }
+
+        // Trim a recognised trailing tag; bail if there is nothing to trim.
+        let len = len_without_trailing_tag(&buf);
+        if len == buf.len() || len < 2 {
+            return None;
+        }
+        buf.truncate(len);
+
+        // Only accept the recovery if what remains is a self-consistent frame — i.e. its trailing
+        // two bytes are the CRC-16 of everything before them. This distinguishes a genuine trailing
+        // tag from a truncated final frame (which must still be reported as an error).
+        let frag = Fragment::new(buf.into_boxed_slice());
+        if !frag.crc_match {
+            return None;
+        }
+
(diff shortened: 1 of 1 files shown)


symphonia-bundle-flac/src/parser.rs
```diff
@@ -276,12 +276,77 @@ impl PacketBuilder {
...
+    /// APEv2 tag). Such a tag gets merged onto the last fragment, breaking its CRC-16 footer so the
+    /// frame is never built. Removing the tag re-exposes the frame's own CRC-16, letting it build.
+    /// Returns `None` when there is nothing to recover or the trimmed data is not a valid frame, in
+    /// which case the caller falls back to its normal truncation handling.
+    fn recover_final_fragment(&mut self) -> Option<Fragment> {
+        if self.frags.is_empty() {
+            return None;
+        }
+
+        // Reassemble the pending fragments into a single buffer.
```
pdeljanov, line 290: This reassemble step should also consider if there is one or more bad fragments.

For example, consider if there are 3 fragments. You could reassemble the fragments as 0+1+2, 1+2, or just 3. In other words, attempt reassembly starting at each fragment This matches the behaviour of `try_build`. Of course, this is not testing all permutations. However, so far this behaviour has been sufficient. My recommendation is to go in the order above: starting with all fragments stitched together, then successively removing one from the front each iteration. This optimistically assumes the file is not corrupted and all fragments will be used which is the most probably case.

pdeljanov (changes requested):
Hi @attilagyorffy , I think this approach is reasonable despite having to hard-code ID3v1 and APEv2 cases. However, I don't see any other way given all the constraints.

To make this solution more complete, please take a look at the comment regarding reassembly permutations.

Otherwise, this looks good to me.

FYI: You'll want to rebase to resolve the clippy warnings.

PR #542, opened 2026-08-04
core: Add missing Clone and slice() to audio structs

MelonShooter:
The `Clone` trait is missing from `AudioSlice` along with two generic audio structs: `GenericAudioBuffer` and `GenericAudioSlice`. `AudioSlice` and `GenericAudioSlice` are effectively cloneable anyways using `AudioSlice::slice`, so may as well allow it for anyone who may want it. The `AudioBuffer` struct that backs `GenericAudioBuffer` already implements `Clone`, so for consistency, the corresponding generic enum struct should also implement `Clone`.

Additionally, `slice()` was missing from two generic enum structs (`GenericAudioBufferRef` and `GenericAudioSlice`) whose backing audio structs already implement this function. I added a macro to reduce duplication and facilitate the change.

### DIFF
--- a/symphonia-core/src/audio/generic.rs
+++ b/symphonia-core/src/audio/generic.rs
@@ -14,10 +14,28 @@ use crate::audio::{
 
 use super::{Audio, AudioBuffer, AudioBytes, AudioMut, AudioSlice, AudioSpec};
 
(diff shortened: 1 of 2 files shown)


pdeljanov:
LGTM. Thanks!

PR #543, opened 2026-08-05
feat: add TimeBase duration conversion helpers

### LINKED ISSUE #530: How to compute time in 0.6.0?
I started migrating my codebase to Symfonia 0.6.0 (coming from 0.5.5), but I'm stuck at computing the duration of a track.

I used to do:

```rust
let dur = match track.duration {
    Some(duration) => time_base.calc_time(duration),
    None => analyze_file_duration(format, track_id, time_base)?, // custom function
};
``` 

But this doesn't work now, as the `calc_time` method takes a `Timestamp`, not a `Duration`. And I haven't found a way to convert the `Duration` into a `Timestamp`. The migration guide doesn't seem to mention how to do it either.

On top of that, the `Track::duration` field docs clearly state:

```md
If a timebase is available, this field can be used to calculate the total duration of the
track in seconds by using [`TimeBase::calc_time`] and passing the duration as the argument.
``` 

Which is incorrect given its current signature:

```rust
pub fn calc_time(&self, ts: Timestamp) -> Option<Time>
```

I suppose it's outdated, but I'm not sure what to do now?

mvanhorn:
Add checked and saturating `TimeBase` duration-conversion helpers in `symphonia-core/src/units.rs`, mirroring the existing `calc_time`/`calc_time_saturating` behavior while accepting the full unsigned `Duration` domain without requiring callers to narrow through `Timestamp`. Symphonia 0.6 separates the unsigned `Duration` type from the signed `Timestamp` type, but `TimeBase::calc_time` only accepts timestamps.

Convert zero and a representative fractional duration with both equivalent reduced and unreduced timebases, and verify the resulting `Time` matches the existing timestamp conversion for values shared by both domains; Convert a `Duration` greater than `i64::MAX` with a sufficiently small timebase and verify the checked helper succeeds without narrowing through `Timestamp`.

Fixes #530

AI was used for assistance.

### DIFF
--- a/docs/guides/migration/0p6.md
+++ b/docs/guides/migration/0p6.md
@@ -31,7 +31,15 @@ In addition to the aforementioned restructuring, metadata and audio primitives h
 ### Track
 
 * Timing information that was previously stored in `CodecParameters` (`time_base`, `n_frames`, `start_ts`, `padding`, `delay`) are now stored directly in `Track`.
-* `Track` adds a new timing field: `duration`. While number of frames could also be considered a duration in time-base units, this is incorrect and only valid when the time-base is `1 / sample_rate`. While this is typical for types of audio files, it is not always the case, particularly for MKV files. The number of frames should be considered a unit-less count of decoded frames. For human readable duration, `duration` should be used instead.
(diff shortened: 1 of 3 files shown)


pdeljanov:
Thanks, LGTM!

PR #548, opened 2026-08-13
ogg: Fix infinite loop when seeking a multiplexed stream

baylesj:
Currently, when consuming an Ogg container that has multiple streams, in some cases we can hang forever when consuming packets because if we encounter a packet with a different track_id, we leave it alone. This PR changes behavior to instead consume the packet, unblocking process.

Chugging along on #501


baylesj:
@pdeljanov chugging along on those code review report findings. LMK what you think

